PR #84, opened 2020-03-04
New xml templates

talogan:


jhkennedy:
Released via containerization; test and prod branch depreciated.

PR #87, opened 2020-06-09
Fix RTC GAMMA errors

jhkennedy:
Current version of the container has 3 major problems: 
1. gamma-0, when request, is not being produced, even though the files are all named indicating it is
2. Some granules are failing with a `MemoryError`
3. Some granules are failing with an `error > max_error`

This PR addresses all three. And some additional clean up.

----

**1. gamma-0**
Previous version of `hyp3-rtc-gamma` had a patched version of `mk_geo_radcal` included that was placed ahead of GAMMA's `mk_geo_radcal` on the path. Because this was proprietary code, `mk_geo_radcal` was removed from this repository. *However*, the patches were the mechanism in which gamma-0 output was being produced. 

**Fix:** the patched has been applied directly to our GAMMA distribution, and the patched distribution is downloaded and installed into the docker container. 

---
**2. MemoryError**

hyp3lib previously did not support the new `create_dem_par` UI for polar stereo projections in GAMMA 2019. As of hyp3lib v1.2.1 ( see asfadmin/hyp3-lib#169) , it does. 

**Fix:** we've updated the hyp3lib requirements.

---

**3. error > max_error**

When co-registration failed, it was throwing a `SystemExit` exception instead of a subclass of `Exception`, so fixing blind except linting errors (`except:`) with `except Exception:` actually caused us to miss the the `SystemExit` exception (a subclass of `BaseException` not `Exception` so that when you want to exit due to critical failures, normal exception handling won't catch it) being thrown, and instead of switching to dead reckoning, we just quit. 

**Fix:** Throw a `CoregistrationError` instead of a `SystemExit` when coregistration fails, and catch `CoregistationError`s specifically. 

---

**Additionally:**

* We were writing our own log file while also trying to use `logging`,which requiring a `display` function to handle both the logger and our log file (:eyes:). Instead, just use a local logging context to *also* write to out log file via  the logger. 
* hyp3lib now includes a `gamma_version` function, so let's use it. 
* hyp3lib now raises a `ExecuteError` when `execute` fails, so let's use it.


hyp3_rtc_gamma/check_coreg.py
```diff
@@ -11,159 +11,157 @@
...
+                    rng_error = float(vals[0])
+                    val = tmp[2].strip()
+                    azi_error = float(val)
+                    logging.info(f"Range std dev: {rng_error}  Azimuth std dev: {azi_error}")
+                    error = np.sqrt(rng_error * rng_error + azi_error * azi_error)
+                    logging.info(f"error is {error}")
+                    if error > max_error:
+                        logging.info("error > max_error")
+                        logging.info("std dev is too high, using dead reckoning")
+                        logging.info("Granule failed coregistration")
```
asjohnston-asf, line 124: `logging.warning` might be called for here

jhkennedy: Good call -- `display` was limited to either `logging.info` or `logging.error`, so the switch to the local context logger would allow us to do that.

asjohnston-asf:
`check_coreg()` deserves to be broken into smaller functions.  In particular the `with` statements are open far too long; have to scroll way way up and check indentation if you want to know if they still apply.  Deferring that for future work, though.

jhkennedy:
@asjohnston-asf agreed on `check_coreg()`. Thanks!

PR #101, opened 2020-06-16
Move ESA citation file generation into science code

jhkennedy:
For HyP3 v2, we want the science code to handle generating a final, unzipped, product entirely. 

This moves adding the ESA citation file from the `v1` entrypoint (after the science code has ran) to inside the science code. 

Because `hyp3proclib` is depreciated, and an updated version of `add_esa_citation` is available in `hyp3lib`, let's use that. 

`add_esa_citation` will throw a `hyp3lib.GranuleError` if the input granule isn't a S1 granule (`hyp3proclib.add_citation` would fail silently). We shouldn't need to worry about this because `rtc_sentinel.py` is only called for S1 granules (`legacy_rtc.py` is called for `E1`, `E2`, `R1`, and `J1` granules, or an unrecognized granule exception is thrown for none of the aforementioned types).

### DIFF
--- a/hyp3_rtc_gamma/__main__.py
+++ b/hyp3_rtc_gamma/__main__.py
@@ -30,7 +30,7 @@ from hyp3proclib import (
     zip_dir,
 )
 from hyp3proclib.db import get_db_connection
-from hyp3proclib.file_system import add_citation, cleanup_workdir
+from hyp3proclib.file_system import cleanup_workdir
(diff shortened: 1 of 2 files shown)


jhkennedy:
Closing this as a "won't fix". We don't need to actually create a seperate ESA citation file because the citation is in the README. Closing this keeps the extra ESA citation file out of v2, but keeps it in v1 for now.

PR #102, opened 2020-06-22
have main_v2() zip output directory before uploading to s3

asjohnston-asf:

### DIFF
--- a/tests/test_v2.py
+++ b/tests/test_v2.py
@@ -3,6 +3,7 @@ from pathlib import Path
 
 import pytest
 from botocore.stub import ANY, Stubber
+
 from hyp3_rtc_gamma import __main__ as main
 
 
--- a/tests/test_v2.py
+++ b/tests/test_v2.py
@@ -61,9 +61,9 @@ def test_upload_file_to_s3(tmp_path, s3_stubber):
     }
     s3_stubber.add_response(method='put_object', expected_params=expected_params, service_response={})
 
-    os.chdir(tmp_path)
-    Path('myFile.zip').touch()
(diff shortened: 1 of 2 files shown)


tests/test_v2.py
```diff
@@ -39,3 +50,31 @@ def test_write_netrc_file(tmp_path):
...
+    expected_params = {
+        'Body': ANY,
+        'Bucket': 'myBucket',
+        'Key': 'myFile.zip',
+        'ContentType': 'application/zip',
+    }
+    s3_stubber.add_response(method='put_object', expected_params=expected_params, service_response={})
+
+    os.chdir(tmp_path)
+    Path('myFile.zip').touch()
```
jhkennedy, line 65: You don't need to do the `os.chdir` bit it looks like, or create a new `Path`. I think it'd work better to do something like:

```python
upload_file = tmp_path / 'myFile.zip'
upload_file.touch()
 main.upload_file_to_s3(upload_file, 'myBucket')
```
Possible need to wrap `upload_file` in `str` if Boto3 doesn't understand pathlib paths

tests/test_v2.py
```diff
@@ -39,3 +50,31 @@ def test_write_netrc_file(tmp_path):
...
+    expected_params = {
+        'Body': ANY,
+        'Bucket': 'myBucket',
+        'Key': 'myPrefix/myFile.txt',
+        'ContentType': 'text/plain',
+    }
+    s3_stubber.add_response(method='put_object', expected_params=expected_params, service_response={})
+
+    os.chdir(tmp_path)
+    Path('myFile.txt').touch()
```
jhkennedy, line 79: Same here

PR #104, opened 2020-06-24
added retry policy to downloads to increase fault tolerence

Jlrine2:
This adds a retry policy to the download file function so that minor http errors and faults might be resolved automatically.

### DIFF
--- a/hyp3_rtc_gamma/__main__.py
+++ b/hyp3_rtc_gamma/__main__.py
@@ -41,7 +41,6 @@ import hyp3_rtc_gamma
 from hyp3_rtc_gamma.rtc_sentinel import rtc_sentinel_gamma
 
 # v2 constants
-CHUNK_SIZE = 5242880
 DATAPOOL_URL = 'https://datapool.asf.alaska.edu'
 EARTHDATA_LOGIN_DOMAIN = 'urs.earthdata.nasa.gov'
 S3_CLIENT = boto3.client('s3')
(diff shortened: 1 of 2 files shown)


hyp3_rtc_gamma/__main__.py
```diff
@@ -96,7 +98,14 @@ def get_download_url(granule):
 def download_file(url):
```
asjohnston-asf, line 98: should probably move towards making this more library-ish:
`def download_file(url, retries=3, backoff=10, chunk_size=5242880):`

hyp3_rtc_gamma/__main__.py
```diff
@@ -96,7 +98,14 @@ def get_download_url(granule):
...
     print(f"\nDownloading {url}")
     local_filename = url.split("/")[-1]
-    with requests.get(url, stream=True) as r:
+    session = Session()
+    retries = Retry(
+        total=3,
+        backoff_factor=10,
+        status_forcelist=[429, 500, 503, 504],
+    )
+    session.mount('https://', HTTPAdapter(retries=retries))
```
asjohnston-asf, line 107: might as well mount http as well:
`session.mount('http://', HTTPAdapter(retries=retries))`
or a single, shorter prefix, but this doesn't seem to be how people do it:
`session.mount('http', HTTPAdapter(retries=retries))`

setup.py
```diff
@@ -37,7 +37,7 @@
         'importlib_metadata',
         'lxml',
         'numpy',
-        'scipy',
+        'scipy', 'requests', 'boto3', 'urllib3'
```
asjohnston-asf, line 40: I slightly prefer one package per line

hyp3_rtc_gamma/__main__.py
asjohnston-asf: or pull the scheme out of `url`, but that seems overkill

setup.py
jhkennedy, line 40: These should be on their own line and sorted -- if/when we start adding restrictions it will be a lot clearer that way

hyp3_rtc_gamma/__main__.py
asjohnston-asf: and turns out it's `max_retries=retries`, my bad on the typo

PR #106, opened 2020-06-25
refactor out_name handling in rtc_sentinel

asjohnston-asf:
No functional changes yet, just eliminating redundant variables related to the product name before actually implementing any changes.

### DIFF
--- a/hyp3_rtc_gamma/rtc_sentinel.py
+++ b/hyp3_rtc_gamma/rtc_sentinel.py
@@ -720,7 +720,7 @@ def rtc_sentinel_gamma(in_file,
     log_file = logging.getLogger().handlers[0].baseFilename
     rtc_name = out_name + "_" + pol + ".tif"
     gamma_ver = gamma_version()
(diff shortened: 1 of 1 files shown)


hyp3_rtc_gamma/rtc_sentinel.py
```diff
@@ -38,6 +38,18 @@
 from hyp3_rtc_gamma.xml2meta import sentinel2meta
 
 
+def get_product_name(granule_name, resolution=30, gamma0=True, power=True, filtered=False):
+    platform = granule_name[0:3]
+    beam_mode = granule_name[4:6]
+    datetime = granule_name[17:32]
+    g = 'g' if gamma0 else 's'
+    p = 'p' if power else 'a'
+    f = 'f' if filtered else 'n'
```
Jlrine2, line 47: I think I prefer parentheses in conditional assignment `result = (true_value if condition else false_value)`

jhkennedy: > I think I prefer parentheses in conditional assignment `result = (true_value if condition else false_value)`
> 

PyCharm and pylint will give "redundant parentheses" warnings for that

PR #114, opened 2020-07-11
refactor create_browse_images

asjohnston-asf:
Eliminate creation of unused `dem`, `ls_map`, and `inc_map` browse files that were being explicitly deleted anyway.

Use NamedTemporaryFile to ensure intermediate `_rgb.tif` and `_sigma.tif` files are deleted.

### DIFF
--- a/hyp3_rtc_gamma/rtc_sentinel.py
+++ b/hyp3_rtc_gamma/rtc_sentinel.py
@@ -7,6 +7,7 @@ import os
 import shutil
 import sys
 import zipfile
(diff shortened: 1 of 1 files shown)


asjohnston-asf:
Turns out the `.png` files for all tifs (including the dem/inc_map/ls_map) are being used to generate the thumbnail images inside the arcgis xml file:  https://github.com/asfadmin/hyp3-rtc-gamma/blob/browse-cleanup/hyp3_rtc_gamma/create_metadata.py#L118

PR #125, opened 2020-07-15
Implement new product naming scheme

asjohnston-asf:

### DIFF
--- a/hyp3_rtc_gamma/__main__.py
+++ b/hyp3_rtc_gamma/__main__.py
@@ -2,6 +2,7 @@
 rtc_gamma processing for HyP3
 """
 import glob
+import logging
 import os
 import shutil
 import sys
@@ -62,7 +63,7 @@ def entry():
 def write_netrc_file(username, password):
     netrc_file = os.path.join(os.environ['HOME'], '.netrc')
     if os.path.isfile(netrc_file):
-        print(f'WARNING - using existing .netrc file: {netrc_file}')
+        logging.warning(f'Using existing .netrc file: {netrc_file}')
     else:
         with open(netrc_file, 'w') as f:
             f.write(f'machine {EARTHDATA_LOGIN_DOMAIN} login {username} password {password}')
@@ -79,7 +80,7 @@ def upload_file_to_s3(path_to_file, bucket, prefix=''):
     key = os.path.join(prefix, os.path.basename(path_to_file))
     extra_args = {'ContentType': get_content_type(key)}
 
-    print(f'Uploading s3://{bucket}/{key}')
+    logging.info(f'Uploading s3://{bucket}/{key}')
     S3_CLIENT.upload_file(path_to_file, bucket, key, extra_args)
 
 
@@ -110,6 +111,9 @@ def main_v2():
     parser.add_argument('--bucket-prefix', default='')
     parser.add_argument('granule')
     args = parser.parse_args()
+
+    logging.basicConfig(format='%(asctime)s - %(levelname)s - %(message)s',
+                        datefmt='%m/%d/%Y %I:%M:%S %p', level=logging.INFO)
 
     write_netrc_file(args.username, args.password)
 
--- a/hyp3_rtc_gamma/__main__.py
+++ b/hyp3_rtc_gamma/__main__.py
@@ -204,13 +204,23 @@ def find_and_remove(dir_, s):
                 os.remove(filepath)
 
 
-def find_product(dir_):
+def find_product_dir(dir_):
     for subdir, dirs, files in os.walk(dir_):
         for d in dirs:
             if d == "PRODUCT":
                 return os.path.join(subdir, d)
 
     return None
+
+
+def find_product_name(dir_):
+    files = glob.glob(f'dir_/**')
+    readme_file = None
+    for file in files:
+        if file.endswith('.README.txt'):
+            readme_file = file
+            break
+    return readme_file.split('.')[0]
 
 
 def download(cfg, granule):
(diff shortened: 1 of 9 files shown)


hyp3_rtc_gamma/__main__.py
```diff
@@ -214,6 +212,16 @@ def find_product(dir_):
...
 
 
+def find_product_name(dir_):
+    files = glob.glob(f'{dir_}/**')
+    readme_file = None
+    for file in files:
+        if file.endswith('.README.txt'):
+            readme_file = file
+            break
+    return readme_file.split('.')[0]
```
asjohnston-asf, line 222: This implementation is pretty lazy and doesn't have tests.  Could refactor all 5 `find_...` functions, they're mostly copy/paste of each other, but I'm not feeling up to it right now.

hyp3_rtc_gamma/rtc_sentinel.py
```diff
@@ -38,15 +39,28 @@
...
-    g = 'g' if gamma0 else 's'
+
+    if orbit_file is None:
+        o = 'O'
+    elif 'POEORB' in orbit_file:
+        o = 'P'
+    elif 'RESORB' in orbit_file:
+        o = 'R'
+    else:
+        o = 'O'
```
asjohnston-asf, line 55: Open to suggestions here, not in love with this implementation.

hyp3_rtc_gamma/rtc_sentinel.py
```diff
@@ -733,7 +747,7 @@ def rtc_sentinel_gamma(in_file,
     create_consolidated_log(out_name, lo_flag, dead_flag, match_flag, gamma_flag, roi,
                             shape, pwr_flag, filter_flag, pol, looks, log_file, smooth, terms,
                             no_cross_pol, par)
-    return 'PRODUCT'
+    return 'PRODUCT', out_name
```
asjohnston-asf, line 750: Both v1 and v2 entrypoints rename the PRODUCT dir to the product name.  I'd like to see that happen here and just return the one value, but I don't feel like digging through the v1 post-processing code to see if anything is hard-coded to look for PRODUCT/.

hyp3_rtc_gamma/etc/README_RTC_GAMMA.txt
```diff
@@ -5,20 +5,24 @@ This folder contains radiometric terrain corrected (RTC) products and their asso
...
-The source granule used to generate the products contained in this folder is:  
-[GRAN_NAME]  
+The folder and each of its contents all share the same base name, using the following convention:
+S1x_yy_aaaaaaaaTbbbbbb_ppo_RTCzz_G_defkl_ssssss
+x:          Sentinel-1 Mission (A or B)
+yy:         Beam Mode
+aaaaaaaa:   Start Date of Acquisition (YYYYMMDD)
+bbbbbb:     Start Time of Acquisition (HHMMSS)
+pp:         Polarization
+o:          Orbit Type (Precise, Restituted, or Predicted)
```
asjohnston-asf, line 15: I don't like using `O` for `Predicted`, they don't feel connected to me at all, I'll follow up with Tom.

hyp3_rtc_gamma/rtc_sentinel.py
Jlrine2: I agree, but nothing much better comes to mind
```python
replacements = {
    'POEORB': 'P',
    'RESORB': 'R',
    'default': 'O'
}
if orbit_file is not None:
    o = [v for k, v in replacements if k in orbit_file][0]
if o is not None:
    o = 'O'
```
for example is much worse

hyp3_rtc_gamma/rtc_sentinel.py
Jlrine2: I think we are making a clear improvement, might be good to push on and document these wants

hyp3_rtc_gamma/__main__.py
jhkennedy: We'll do some cleanup and moving stuff into hyp3lib as we start adding more processes, so we can circle back to this then

hyp3_rtc_gamma/__main__.py
```diff
@@ -346,20 +352,18 @@ def process_rtc_gamma(cfg, n):
...
-                find_and_remove(product, '_dem.tif')
+                find_and_remove(product_dir, '_dem.tif')
             if extra_arg_is(cfg, "include_inc", "no"):
-                find_and_remove(product, '_inc_map.tif')
+                find_and_remove(product_dir, '_inc_map.tif')
 
-            out_name = build_output_name(g, cfg['workdir'], opts_str + cfg['suffix'])
-            if "STD_L0_F" in out_name:
-                out_name = out_name.replace("_L0", "")
+            out_name = find_product_name(product_dir)
```
asjohnston-asf, line 366: need to address non-sentinel products, I think the other use case is PALSAR via rtc_legacy.py

jhkennedy: I think other products will need to be addressed separately; there is a bit more work to do from what I can tell to get them supported in the containers (missing dependencies). Maybe just make an issue for this

asjohnston-asf: Yeah, should just get confirmation from Kirk that things are already broken in production, in which case this change won't make things any worse.

asjohnston-asf:
container `2.0.7.dev20_g5992e11` passes a happy path test for v1 against the hyp3-test database

asjohnston-asf:
found a pathing problem applying slc orbit files, get_orb downloads to the root working directory, but applying the orbit file happens in a subdirectory at https://github.com/ASFHyP3/hyp3-lib/blob/develop/hyp3lib/par_s1_slc_single.py#L68 , so the orbit file isn't found

hyp3_rtc_gamma/rtc_sentinel.py
```diff
@@ -657,6 +657,8 @@ def rtc_sentinel_gamma(in_file,
     report_kwargs(in_file, out_name, res, dem, roi, shape, match_flag, dead_flag, gamma_flag, lo_flag,
                   pwr_flag, filter_flag, looks, terms, par, no_cross_pol, smooth, area, orbit_file)
 
+    orbit_file = os.path.join(os.getcwd(), orbit_file)  # ingest_S1_granule requires absolute path
```
jhkennedy, line 660: might be more explicit to do an `os.path.abspath`

jhkennedy: and could probably me moved into your helper `fetch_orbit_file` function

hyp3_rtc_gamma/etc/README_RTC_GAMMA.txt
jhkennedy: Sounds like, U for Unknown would be a better letter as we're not downloading the Predicted files (unless that's also included in the SAFE and we're pulling it from there)

hyp3_rtc_gamma/rtc_sentinel.py
asjohnston-asf: I prefer handing just the filename to `get_product_name` and `report_kwargs` (for now).  Implemented `os.path.abspath()`.

PR #141, opened 2020-07-27
refactor v1 entrypoint

asjohnston-asf:
- removes `This is an RTC product from 2015-06-21. You are receiving this product 1863 days, 9:23:09 after it was acquired.` from the bottom of email notifications
- removes `ESA_citation.txt` from output zip file
- input scenes are now downloaded directly from ngap distribution endpoint rather than datapool
- removes non-functional support for non-S1 products
- removes inaccessible clipping option (option is still available via command line)
- removes support for `Flatten Terrain` aka `keep_area` aka `area` option that's only accessible on the test website (option is still available via command line)


hyp3_rtc_gamma/__main__.py
```diff
@@ -134,251 +124,82 @@ def main_v2():
...
-
-                upload_product(zip_file, cfg, conn, browse_path=browse_path)
-                success(conn, cfg)
+        logging.basicConfig(format='%(asctime)s - %(levelname)s - %(message)s',
+                            datefmt='%m/%d/%Y %I:%M:%S %p', level=logging.INFO)
+        logging.info(f'Processing GAMMA RTC "{cfg["sub_name"]}" for "{cfg["username"]}"')
+
+        granule = cfg['granule']
+        if not re.match('S1[AB]_.._[SLC|GRD]', granule):
+            raise ValueError(f'Invalid granule, only S1 SLC and GRD data are supported: {granule}')
```
jhkennedy, line 156: This could be a `hyp3lib.GranuleError` but `ValueError` is also fine

hyp3_rtc_gamma/__main__.py
```diff
@@ -134,251 +124,82 @@ def main_v2():
...
+            find_and_remove(product_dir, '*_dem.tif*')
+        if extra_arg_is(cfg, 'include_inc', 'no'):
+            find_and_remove(product_dir, '*_inc_map.tif*')
+
+        zip_file = make_archive(base_name=product_dir, format='zip', base_dir=product_dir)
+        cfg['final_product_size'] = [os.stat(zip_file).st_size, ]
+        cfg['attachment'] = find_png(product_dir)
+
+        for handler in log.handlers:
+            log.removeHandler(handler)
```
jhkennedy, line 190: Should this be up where you setup the logging? 
https://github.com/ASFHyP3/hyp3-rtc-gamma/pull/141/files#diff-8be91576c32694692cf90f1752b8c91fR150-R152

PR #153, opened 2020-08-10
add file association tag

Jlrine2:


jhkennedy:
I believe we've decided that this implementation isn't how we'd like to go about it, and it's getting rapidly outdated at this point. 

Closing.

PR #159, opened 2020-08-28
remove obsolete scipy dependency

asjohnston-asf:


Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN apt-get update && apt-get upgrade -y && \
     libgtk2.0-bin libgtk2.0-common libgtk2.0-dev \
     libhdf5-100 libhdf5-dev liblapack-dev liblapack3 \
     netpbm python3-dev python3-h5py python3-netcdf4 \
-    python3-matplotlib python3-pip python3-scipy tcsh unzip vim wget && \
```
jhkennedy, line 30: We'll still need this for hyp3-lib

PR #178, opened 2020-10-23
draft of --include-scattering-area

asjohnston-asf:
Removes code for producing flattened backscatter tifs introduced in https://github.com/ASFHyP3/hyp3-rtc-gamma/commit/ea08afac67c34c0f6064ccc57f586cc4bd03096c

Replaces the existing `--area` parameter with `--include-scattering-area`, which results in a `_area.tif` being included in the output product directory and/or zip file.

This approach avoids further patching `mk_geo_radcal` by waiting to generate the area map until after the co-pol call to `mk_geo_radcal` mode 3 has completed.

Product readme in hyp3-metadata-templates should be updated to include a description of the additional output file.

Need to confirm with Kirk whether hyp3-metadata-templates should also create an ArcGIS-compatible .xml metadata file for the area tif, and if that should include a thumbnail image.

Product guide in ASFHyP3 should be updated to reflect the additional output file.

hyp3 api and orchestration should be updated to support `include_scattering_area` as an `RTC_GAMMA` job parameter.

hyp3-testing golden set should be updated to exercise the `include_scattering_area` parameter

Need to confirm with Kirk that `include-scattering-area` will *not* be made available in HyP3 v1.

*thing I forgot here*

### DIFF
--- a/hyp3_rtc_gamma/rtc_sentinel.py
+++ b/hyp3_rtc_gamma/rtc_sentinel.py
@@ -155,7 +155,7 @@ def reproject_dir(dem_type, res, prod_dir=None):
 
 
 def report_kwargs(in_name, out_name, res, dem, roi, shape, match_flag, dead_flag, gamma_flag,
-                  pwr_flag, filter_flag, looks, terms, par, no_cross_pol, smooth, area, orbit_file):
+                  pwr_flag, filter_flag, looks, terms, par, no_cross_pol, smooth, include_area_map, orbit_file):
     logging.info("Parameters for this run:")
     logging.info("    Input name                        : {}".format(in_name))
     logging.info("    Output name                       : {}".format(out_name))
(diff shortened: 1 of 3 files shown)


hyp3_rtc_gamma/rtc_sentinel.py
```diff
@@ -369,19 +352,24 @@ def process_2nd_pol(in_file, rtc_name, cpol, res, look_fact, gamma_flag, filter_
...
 
     os.chdir(home_dir)
 
 
+def create_area_map(mli_par, dem_seg_par, input_pix, input_map, output_name):
+    mli_width = getParameter(mli_par, 'range_samples')
+    dem_width = getParameter(dem_seg_par, 'width')
+    dem_lines = getParameter(dem_seg_par, 'nlines')
+
+    execute(f'geocode_back {input_pix} {mli_width} {input_map} area_map.pix {dem_width} {dem_lines} 2 0', uselogging=True)
```
asjohnston-asf, line 369: opportunities to use `tempfile.NamedTemporaryFile` and `subprocess.run` here

jhkennedy (approved):

PR #187, opened 2020-11-17
use mk_geo_radcal2 to remove no data values

talogan:
This PR requires changes to `GAMMA_SOFTWARE-20191203/DIFF/scripts/mk_geo_radcal2`.
Removes no data values from lakes and rivers.


hyp3_rtc_gamma/rtc_sentinel.py
```diff
@@ -595,7 +595,7 @@ def rtc_sentinel_gamma(in_file,
                 browse_res, dem, terms, par=par, orbit_file=orbit_file)
 
     if include_scattering_area:
-        create_area_geotiff(f'geo_{pol}/image_1.pix', f'geo_{pol}/image_1.map_to_rdc', f'{out_name}.{pol}.mgrd.par',
+        create_area_geotiff(f'geo_{pol}/image_gamma0_1.pix', f'geo_{pol}/image_1.map_to_rdc', f'{out_name}.{pol}.mgrd.par',
```
asjohnston-asf, line 598: Is this change related to `mk_geo_radcal2`?

asjohnston-asf: Ah, I think I see it, we'll likely need to reference our radiometry flag to know which image to pull here.  It looks like the patch you posted is only the additional changes, not including the pre-existing patch to create the -g flag?

talogan: When I firs tran the code with mk_geo_radcal2, I got an error that image_1.pix was not found.  I traced it down in mk_geo_radcal2 and found that the file is now called image_gamma0_1.pix.  So, I changed the name

jhkennedy (commented):
I was able to pull down this container and run our "favorite" granule through the HyP3v2 entrypoint. 

Visual inspection between this and develop shows the same RTC product, but the image geo-reference is different (slight shift). `gdalcompare.py` gives:

```
gdalcompare.py S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_4411/S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_4411_VV.tif S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_EF9C/S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_EF9C_VV.tif
Files differ at the binary level.
GeoTransforms Differ:
  Golden: (365250.0, 30.0, 0.0, 1647840.0, 0.0, -30.0)
  New:    (365730.0, 30.0, 0.0, 1647360.0, 0.0, -30.0)
Band size mismatch (band=0 golden=[8933,4444], new=[8901,4412])
Differences Found: 3
```
Above: new == this branch; Golden == develop

talogan:
The test results are counter intuitive.  The Golden test is larger by 32
pixels in each dimension.  Presumably this is 16 pixels in each direction,
which works out with the corners being off by 480 meters (32 pixels is 960
meters half of which is 480 meters).

But, what was supposed to happen was that the image would get BIGGER, not
smaller.

So, I'm pretty confused by this result.  I'll look into it.

jhkennedy:
Reruning with hyp3lib v1.6.3, I'm still seeing a different geotransform:

```
gdalcompare.py S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_4411/S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_4411_VV.tif S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_6443/S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_6443_VV.tif
Files differ at the binary level.
GeoTransforms Differ:
  Golden: (365250.0, 30.0, 0.0, 1647840.0, 0.0, -30.0)
  New:    (365730.0, 30.0, 0.0, 1647360.0, 0.0, -30.0)
Band size mismatch (band=0 golden=[8933,4444], new=[8901,4412])
Differences Found: 3
```
Above: new == HEAD of this branch; Golden == develop

Comparing this branch without the new hyp3lib (golden) with the HEAD of this branch (new), gives:

```
gdalcompare.py S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_EF9C/S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_EF9C_VV.tif S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_6443/S1A_IW_20150621T120220_SVP_RTC30_G_gpuned_6443_VV.tif
Files differ at the binary level.
Differences Found: 1
```

So, the hyp3lib version doesn't seem to matter here (I did confirm the container is using the new one) -- probably because it's an SLC.

jhkennedy (approved):
I think since we're able to run everything, we can merge into develop so we can easily run jobs and investigate the geotransform and verify the changes.

PR #195, opened 2020-12-17
Remove --roi, --shape, --smooth parameters from rtc_sentinel.py

asjohnston-asf:
This PR proposes to remove several unused options for preparing DEM data:
- `--roi`, used to subset the output of `hyp3lib.get_dem` to a bounding box before processing.
- `--shape`, used to subset the output of `hyp3lib.get_dem` to the extent of a shapefile before processing.
- `--smooth`, used to run GAMMA's `fill_gaps` method on REMA DEM tiles before processing (REMA is not currently in use).

These simplifications clear the way for further DEM-related improvements scheduled for Q1 2021.

Thoughts?  Has anyone leveraged these options in the last, say, 6 months?


talogan:
The ability to cut to a shape file or a roi is very useful when dealing
with specific study areas.
I definitely would be against getting rid of those options in the science
code.
What harm are they doing?

asjohnston-asf:
No specific harm, but they're adding inertia and creating more surface area to test which can slow down future development, so I'd rather retire them if they're not being actively used.

I was thinking we'd leave the `--external-dem` option, so you could provide a pre-subsetted DEM on the off chance you want one.  We'll need subsetting options down the road if/when we make a user-facing priority, but that hasn't been on the roadmap in any of our epic discussions, so I'd rather keep things lean for now and come back to it fresh later.

Alternatively, how would you feel about dropping `--shape` (the more complicated one) but leaving `--roi`?

PR #198, opened 2021-01-07
capture both co-pol and cross-pol mk_geo_radcal mode 3 log files

asjohnston-asf:


jhkennedy (approved):
I think generally the appended log files will show up alphabetically, but if we don't see that happening, we might want to circle back and enforce that.

PR #207, opened 2021-02-05
Add --include-rgb option for RTC

asjohnston-asf:
Should we log something if the user selects `--include-rgb` for a single-pol granule?

As written, the rgb browse is the only browse generated outside of `create_browse_images`.  I could be convinced to move it, but I thought it was easier to do all at once while all the file paths and conditionals are handy.

`hyp3lib.area2point.fix_geotiff_locations` only works for single-band geotiffs, so I've re-implemented it as `hyp3_gamma.util.rtc_sentinel.set_pixel_as_point`.  Arguably that should be in `hyp3lib` but at the moment rtc is the only place in our code base where `fix_geotiff_locations` is invoked.  All the .tif files already have `AREA_OR_POINT=Point` by the time they reach this step, though, so I'll need to sync up with Tom on whether we actually need to take action or not.  The rgb tif is the only `AREA_OR_POINT=Area` tif at that stage.

`hyp3lib.rtc2color.rtc2color()` doesn't assign a NoData value to the rgb tiff; https://github.com/ASFHyP3/hyp3-lib/pull/241 corrects that

Metadata updates are outside the scope of this PR:
- update readme
- add rgb.tif.xml ArcGIS file
- update rgb.png.xml file?
- update product guide?


CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+* An `include_rgb` option has been added to `rtc_sentinel.py` to include an RGB decomoposition GeoTIFF in the output
```
jhkennedy, line 12: should probably write this like "An `--include-rgb` option..." since you're talking about the entrypoint, might be worth  calling out the funtion parameter as well. Something like:

```
* Ability to include an RGB deompostion GeoTIFF in the output
  * an `--include-rgb` option has been added to `rtc_sentinel.py`
  * an `include_rgb` keyword argument has been added to `hyp3_gamma.rtc.rtc_sentinel.rtc_sentinel_gamma`
```

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+  package when processing a dual-polarization product.
+
+### Changed
+* Color browse images are only generated for RTC products when the `include_rgb` option is selected.
```
jhkennedy, line 16: maybe just say
```
* Color browse images are only generated for RTC products when requested
```

jhkennedy (changes requested):
Just some changelog language changes; otherwise looks good

asjohnston-asf:
`fix_geotiff_locations` throws away everything that's not band 1 at the moment, so we get a good browse image but only the red band in the geotiff

jhkennedy (approved):

PR #222, opened 2021-03-09
Initial implementation of RTC via Copernicus DEM

asjohnston-asf:
`cop30.geojson` has one feature per DEM tile.  Each feature has a `file_path` property with the full `/vsicurl/https://...` path to the GeoTIFF in the `copernicus-dem-30m` S3 bucket.

The Copernicus DEM has no `NoData` value.  Ocean areas have a data value of zero.  GAMMA's `dem_import` assumes `NoData Value=0` if one is not specified in the input DEM GeoTIFF.  I'm currently using the `zflg` parameter to interpret data values of zero as "close to zero, not `NoData`".  If we could pad the input DEM with zeros to the full extent of the granule (including areas with no dem coverage) then I think we'd get "geocode only" behavior for areas with no DEM coverage.

Not currently subsetting the DEM to the extent of the granule.  GAMMA will still subset the dem segment to the extent of the granule.

Metadata updates are pending, see https://github.com/ASFHyP3/hyp3-metadata-templates/pull/50

Legacy DEM handling is currently removed.  Adding an option to use the legacy SRTM/NED DEMs is still pending.  The --dem and --bbox command line options can also be re-implemented if needed.

Some (all?) `hyp3_gamma.dem` methods could be migrated to `hyp3lib`.  At a minimum, `GDALConfigManager` should be part of a different module.

DEM coverage geojson is currently hardcoded against an S3 location.  May revisit where to keep the canonical copy, and how/if to provide a custom geojson.


hyp3_gamma/dem.py
```diff
@@ -0,0 +1,115 @@
...
+from osgeo import gdal, ogr
+
+DEM_GEOJSON = '/vsicurl/https://asf-dem-west.s3.us-west-2.amazonaws.com/v2/cop30.geojson'
+
+gdal.UseExceptions()
+ogr.UseExceptions()
+
+
+def get_geometry_from_kml(kml_file: str) -> ogr.Geometry:
+    # TODO suppress warnings about Date parsing?
```
asjohnston-asf, line 18:
```
Warning 1: The output driver does not natively support DateTime type for field timestamp. Misconversion can happen. -mapFieldType can be used to control field type conversion.
Warning 1: The output driver does not natively support DateTime type for field begin. Misconversion can happen. -mapFieldType can be used to control field type conversion.
Warning 1: The output driver does not natively support DateTime type for field end. Misconversion can happen. -mapFieldType can be used to control field type conversion.
```

hyp3_gamma/dem.py
```diff
@@ -0,0 +1,130 @@
...
+def get_geometry_from_kml(kml_file: str) -> ogr.Geometry:
+    response = run(['ogr2ogr', '-wrapdateline', '-datelineoffset', '20', '-f', 'GeoJSON', '-mapfieldtype',
+                    'DateTime=String', '/vsistdout', kml_file], stdout=PIPE, check=True)
+    geojson = json.loads(response.stdout)
+    geometry = json.dumps(geojson['features'][0]['geometry'])
+    return ogr.CreateGeometryFromJson(geometry)
+
+
+def get_dem_features() -> Generator[ogr.Feature, None, None]:
+    ds = ogr.Open(DEM_GEOJSON)
```
Jlrine2, line 25: this might not get properly closed, if `ds` has a method `ds.close` this can be resolved with
```py
from contextlib import closing
def get_dem_features() -> Generator[ogr.Feature, None, None]:
    ds = ogr.Open(DEM_GEOJSON)
    with closing(ds):
        ....
```
alternatively, we can write out own context manager
```py
@context_manager
def ogr_file_managed(file_path):
    ds = ogr.Open(file_path)
    try:
        yield ds
    finally:
        del ds

def get_dem_features() -> Generator[ogr.Feature, None, None]:
    with ogr_file_managed(DEM_GEOJSON) as ds:
        ....
```

Jlrine2: Nevermind, `ds` will be cleaned up when teh function that calls it goes out of scope

jhkennedy: GDAL shit is weird; you have to dereference to close. 

```python
from osgeo import ogr
DEM_GEOJSON = '/vsicurl/https://asf-dem-west.s3.us-west-2.amazonaws.com/v2/cop30.geojson'
ds = ogr.Open(DEM_GEOJSON)
ds.close()
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/home/jhkennedy/mambaforge/envs/hyp3-autorift/lib/python3.8/site-packages/osgeo/ogr.py", line 452, in <lambda>
    __getattr__ = lambda self, name: _swig_getattr(self, DataSource, name)
  File "/home/jhkennedy/mambaforge/envs/hyp3-autorift/lib/python3.8/site-packages/osgeo/ogr.py", line 80, in _swig_getattr
    raise AttributeError("'%s' object has no attribute '%s'" % (class_type.__name__, name))
AttributeError: 'DataSource' object has no attribute 'close'
```

You can implement a context manager, and that's what [rasterio does](https://github.com/mapbox/rasterio/blob/master/rasterio/__init__.py#L43), but [it's complicated](https://stackoverflow.com/questions/49648360/use-with-statement-to-close-gdal-datasets) to do well for all GDAL things. 

If we want things like context mangers and what not, we should use [rasterio  instead](https://rasterio.readthedocs.io/en/latest/topics/switch.html) of vanilla gdal, but [it's all in or nothing](https://rasterio.readthedocs.io/en/latest/topics/switch.html#mutual-incompatibilities:~:text=Rasterio%20and%20GDAL%E2%80%99s%20bindings%20can%20contend,of%20import%20osgeo.gdal%20or%20import%20rasterio.) and we do enough with GDAL that switching would be a big lift.

hyp3_gamma/dem.py
```diff
@@ -0,0 +1,126 @@
...
+DEM_GEOJSON = '/vsicurl/https://asf-dem-west.s3.amazonaws.com/v2/cop30.geojson'
+
+gdal.UseExceptions()
+ogr.UseExceptions()
+
+
+def get_geometry_from_kml(kml_file: str) -> ogr.Geometry:
+    response = run(['ogr2ogr', '-wrapdateline', '-datelineoffset', '20', '-f', 'GeoJSON', '-mapfieldtype',
+                    'DateTime=String', '/vsistdout', kml_file], stdout=PIPE, check=True)
+    geojson = json.loads(response.stdout)
```
jhkennedy, line 19: I'm not sure I like `response` as the variable name holding a `CompletedProcess` object, and I think I'd prefer to keep the subprocess namespace since `run` is pretty generic.  

This might look better like:
```python
    ogr2ogr_cmd = ['ogr2ogr', '-wrapdateline', '-datelineoffset', '20', '-f', 'GeoJSON', '-mapfieldtype', 'DateTime=String', '/vsistdout', kml_file]
    geojson = run(ogr2ogr, stdout=PIPE, check=True).stdout
    geometry = json.loads(geojson)['features'][0]['geometry']
return ogr.CreateGeometryFromJson(json.dumps(geometry))

Jlrine2 (approved):

PR #223, opened 2021-03-15
Proceed without precise orbits

jhkennedy:

### DIFF
--- a/hyp3_gamma/insar/ifm_sentinel.py
+++ b/hyp3_gamma/insar/ifm_sentinel.py
@@ -277,10 +277,10 @@ def insar_sentinel_gamma(reference_file, secondary_file, rlooks=20, alooks=4, lo
             log.warning(f'Proceeding using original predicted orbit data included with {granule}')
             orbit_file = None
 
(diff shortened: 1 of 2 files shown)


hyp3_gamma/insar/ifm_sentinel.py
```diff
@@ -270,9 +270,18 @@ def insar_sentinel_gamma(reference_file, secondary_file, rlooks=20, alooks=4, lo
...
+        try:
+            orbit_file, _ = downloadSentinelOrbitFile(granule)
+        except OrbitDownloadError as e:
+            log.warning(e)
+            log.warning(f'Proceeding using original predicted orbit data included with {granule}')
+            orbit_file = None
+
+        if orbit_file is not None:
+            orbit_file = os.path.abspath(orbit_file)
+            orbit_files.append(orbit_file)
```
asjohnston-asf, line 282: we want `orbit_files` to be `(None, None)` rather than `[]` when it's passed to `get_product_name()` and `least_precise_orbit_of()`

asjohnston-asf (approved):

PR #232, opened 2021-03-29
Add dem_name option for RTC supporting copernicus or legacy

asjohnston-asf:
`--dem-name legacy S1A_IW_SLC__1SSV_20150621T120220_20150621T120232_006471_008934_72D8` passes golden testing compared to `main` (with all other parameters as the defaults).

Similarly, `--dem-name copernicus S1A_IW_SLC__1SSV_20150621T120220_20150621T120232_006471_008934_72D8` passes golden testing compared to `develop`.

### DIFF
--- a/hyp3_gamma/rtc/rtc_sentinel.py
+++ b/hyp3_gamma/rtc/rtc_sentinel.py
@@ -13,10 +13,11 @@ from secrets import token_hex
 from tempfile import NamedTemporaryFile, TemporaryDirectory
 
 from hyp3_metadata import create_metadata_file_set
-from hyp3lib import ExecuteError, GranuleError, OrbitDownloadError
+from hyp3lib import DemError, ExecuteError, GranuleError, OrbitDownloadError
 from hyp3lib.byteSigmaScale import byteSigmaScale
 from hyp3lib.createAmp import createAmp
 from hyp3lib.execute import execute
+from hyp3lib.getDemFor import getDemFile
 from hyp3lib.getParameter import getParameter
 from hyp3lib.get_orb import downloadSentinelOrbitFile
 from hyp3lib.makeAsfBrowse import makeAsfBrowse
@@ -24,6 +25,7 @@ from hyp3lib.make_cogs import cogify_dir
 from hyp3lib.raster_boundary2shape import raster_boundary2shape
 from hyp3lib.rtc2color import rtc2color
 from hyp3lib.system import gamma_version
+from hyp3lib.utm2dem import utm2dem
 from osgeo import gdal, gdalconst
 
 import hyp3_gamma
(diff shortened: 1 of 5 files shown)


hyp3_gamma/rtc/rtc_sentinel.py
```diff
@@ -99,6 +101,7 @@ def log_parameters(safe_dir, resolution, radiometry, scale, speckle_filter, dem_
     log.info(f'    Include RGB decomposition : {include_rgb}')
     log.info(f'    Orbit file                : {orbit_file or "Original Predicted"}')
     log.info(f'    Output name               : {product_name}')
+    log.info(f'    DEM name                  : {dem_name}')
```
asjohnston-asf, line 104: `DEM strategy` or something similar might be a better label here?  This is before we've prepared the DEM, so we don't yet know the specific DEM that legacy processing will use.

jhkennedy: Yeah, this was kinda the struggle I was having yesterday  -- I like DEM strategy, but I'm not sure what we'd call the opposite of `legacy`. 

We've got a specify DEM, or "the best of a set", which I can't figure out how to explain well concisely

jhkennedy: Can punt this for a future discussion though

asjohnston-asf: I'm comfortable leaving this as-is, literally echoing the value of the `dem_name` input, which is what all the other lines in this function are doing

hyp3_gamma/rtc/rtc_sentinel.py
```diff
@@ -272,17 +276,23 @@ def rtc_sentinel_gamma(safe_dir: str, resolution: float = 30.0, radiometry: str
...
-    dem_type = 'COP30'
+    dem_tif = 'dem.tif'
     dem_image = 'dem.image'
     dem_par = 'dem.par'
-    geometry = get_geometry_from_kml(f'{safe_dir}/preview/map-overlay.kml')
-    with NamedTemporaryFile() as dem_tif:
-        prepare_dem_geotiff(dem_tif.name, geometry)
-        run(f'dem_import {dem_tif.name} {dem_image} {dem_par} - - $DIFF_HOME/scripts/egm2008-5.dem '
+    if dem_name == 'copernicus':
+        dem_type = 'COP30'
```
jhkennedy, line 286: Since this is only used for `hyp3_metadata`, we should probably just update it to use `copernicus`, though we may need to specify a `dem_resolution` in the future...

No matter for this PR though

jhkennedy (approved):

hyp3_gamma/rtc/rtc_sentinel.py
asjohnston-asf: And with this morning's round of hyp3-metadata-templates edits, the name value is only displayed in the README.md.txt:

```
The DEM used for this product is <dem_name>, which has a native pixel spacing of 1 arc second (about 30 meters).

The Copernicus DEM GLO-30 is a global Digital Surface Model (DSM) derived from the WorldDEM. The WorldDEM is based on radar satellite data acquired by the TanDEM-X mission, which was funded by the German Aerospace Center (DLR) and Airbus Defence and Space, and edited to flatten water bodies, provide consistent flow of rivers, and apply corrections to shore/coastlines and special features. For an overview of the dataset, visit https://spacedata.copernicus.eu/fr/dataset-details?articleId=394198. The product handbook is available at https://spacedata.copernicus.eu/documents/20126/0/GEO1988-CopernicusDEM-SPE-002_ProductHandbook_I1.00.pdf.
```

Heidi and I were thinking `GLO-30` would be the best thing to put there, since `Copernicus DEM GLO-30` appears in the next sentence.  But that'll be a subsequent hyp3-gamma PR after the next hyp3-metadata-templates release.

jhkennedy: makes sense to me; `dem_name` can just be `GLO-30` then I think

PR #233, opened 2021-03-31
restore --dem and --bbox options

asjohnston-asf:
`--dem-name legacy S1A_IW_SLC__1SSV_20150621T120220_20150621T120232_006471_008934_72D8` passes golden testing compared to `main` (with all other parameters as the defaults)

`--bbox -92 14.2 -91.85 14.35 --dem-name legacy S1A_IW_SLC__1SSV_20150621T120220_20150621T120232_006471_008934_72D8.SAFE` passes compared to `main`

`--dem-name copernicus S1A_IW_SLC__1SSV_20150621T120220_20150621T120232_006471_008934_72D8` passes compared to `develop`

`--bbox` with `--dem-name copernicus` doesn't subset to the exact bbox like with `--dem-name legacy`, but will limit the dem mosaic to tiles within 0.15 degrees of the bbox.

`--dem` now feeds the GeoTIFF through `dem_import` with EGM2008 correction and the default NoData parameters.  Still requires a UTM projection (or at least something with a pixel size in meters, not degrees), or processing fails with `mk_geo_radcal2: ERROR calloc_2d: memory allocation error for data values` (since `mk_geo_radcal2` is called with a resolution of `30.0`).

### DIFF
--- a/hyp3_gamma/rtc/rtc_sentinel.py
+++ b/hyp3_gamma/rtc/rtc_sentinel.py
@@ -294,7 +294,7 @@ def rtc_sentinel_gamma(safe_dir: str, resolution: float = 30.0, radiometry: str 
         looks: Number of azimuth looks to take. Will be selected automatically if not specified.  Range and filter looks
             are selected automatically based on azimuth looks and product type.
         skip_cross_pol: Do not include the co-polarization backscatter GeoTIFF in the output package.
-        dem_name: DEM to use for RTC processing; `copernicus` or `legacy`.
+        dem_name: DEM to use for RTC processing; `copernicus` or `legacy`. `dem_name` is ignored if `dem` is provided.
 
     Returns:
         product_name: Name of the output product directory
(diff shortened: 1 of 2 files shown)


hyp3_gamma/rtc/rtc_sentinel.py
```diff
@@ -409,11 +439,18 @@ def main():
...
 
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument('--dem', help='Path to the DEM to use for RTC processing. Must be a GeoTIFF in a UTM projection.'
+                                     ' A DEM will be selected automatically if not provided.')
+    group.add_argument('--dem-name', choices=('copernicus', 'legacy'), default='copernicus',
+                       help='DEM to use for RTC processing.')
+
+    parser.add_argument('--bbox', type=float, nargs=4, metavar=('LON_MIN', 'LAT_MIN', 'LON_MAX', 'LAT_MAX'),
+                        help='Subset the output images to the given lat/lon bounding box. Ignored if --dem is '
+                             'provided.')
```
jhkennedy, line 450: Might be a place for [a mutually exclusive group](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group)...

asjohnston-asf: `--dem` is exclusive to both `--dem-name` and `--bbox`, but I didn't see a way to enforce that with argparse.  Adding `--bbox` to the exclusive group prevents it from being used together with `--dem-name`.

PR #236, opened 2021-04-04
zero-fill Copernicus DEM to extent of input geometry

asjohnston-asf:
This is intended to implement geocoding during RTC in regions where no DEM is available.  Only implemented for when `dem_name='copernicus'`.

Doesn't (yet) play nicely with Copernicus for InSAR (https://github.com/ASFHyP3/hyp3-gamma/pull/235). I imagine InSAR jobs won't want to zero-fill areas where no DEM is available, and still raise an error when no DEM coverage is available at all.

### DIFF
--- a/hyp3_gamma/dem.py
+++ b/hyp3_gamma/dem.py
@@ -4,7 +4,6 @@ from subprocess import PIPE, run
 from tempfile import TemporaryDirectory
 from typing import Generator, List
 
-from hyp3lib import DemError
 from osgeo import gdal, ogr
 
 from hyp3_gamma.util import GDALConfigManager
--- a/hyp3_gamma/dem.py
+++ b/hyp3_gamma/dem.py
@@ -31,17 +31,14 @@ def get_dem_features() -> Generator[ogr.Feature, None, None]:
     del ds
 
 
-def intersects_dem(geometry: ogr.Geometry) -> bool:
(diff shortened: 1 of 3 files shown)


hyp3_gamma/dem.py
```diff
@@ -31,17 +30,14 @@ def get_dem_features() -> Generator[ogr.Feature, None, None]:
...
-
-
 def get_dem_file_paths(geometry: ogr.Geometry) -> List[str]:
     file_paths = []
     for feature in get_dem_features():
         if feature.GetGeometryRef().Intersects(geometry):
             file_paths.append(feature.GetField('file_path'))
+    if not file_paths:
+        file_paths.append('/vsicurl/https://copernicus-dem-30m.s3.amazonaws.com/'
+                          'Copernicus_DSM_COG_10_N00_00_E106_00_DEM/Copernicus_DSM_COG_10_N00_00_E106_00_DEM.tif')
```
asjohnston-asf, line 40: Open to suggestions on how to handle the "no dem tiles" case.  I think we can effectively hand any non-intersecting or zero-filled geotiff/vrt to the gdal.Warp command and get all-zero output.

[gdal_create](https://gdal.org/programs/gdal_create.html) could be a good option if/when we upgrade to gdal>=3.2.0

PR #244, opened 2021-04-22
Adding support for mintPy

talogan:


asjohnston-asf:
I imagine these are the next steps. Any concerns with adding # 1 and # 2 to this PR?
1. Add a description of the Incidence Angle geotiff to the [README](https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/insar/etc/README_InSAR_GAMMA.txt).  Can you cover that, @talogan ?  I assume we'll cover the xml metadata and product guide separately?
1. Add a corresponding parameter to the [HyP3 v2 entrypoint](https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/__main__.py#L87).  I can grab that part if you like.
1. Add a corresponding parameter to the [HyP3 API](https://github.com/ASFHyP3/hyp3/blob/develop/job_types.yml#L78)
1. Add a corresponding parameter to the [HyP3 SDK](https://github.com/ASFHyP3/hyp3-sdk/blob/develop/hyp3_sdk/hyp3.py#L319)
1. Add a corresponding parameter to the Vertex On Demand queue

talogan:
Thanks Andrew - I can add a description to to the README.

hyp3_gamma/insar/etc/README_InSAR_GAMMA.txt
```diff
@@ -58,6 +58,7 @@ The files generated in this process include:
 6. Amplitude Image (GeoTIFF)
 7. Parameter Documentation (Text File)
 8. Look Vector maps (GeoTIFFs) - *Optional*
+9. Incidence angle map (GeoTIFFs) - *Optional*
```
asjohnston-asf, line 61: capital A in `Angle`, and singular `GeoTIFF`?

hyp3_gamma/insar/etc/README_InSAR_GAMMA.txt
```diff
@@ -128,6 +129,14 @@ Two look vector maps are output with the product as GeoTIFFs. The lv_theta map i
 
 *Note that these files are optional. Select the "Angle Maps" option in the HyP3 Processing Options to include them in the product package.*
 
+----------------
+## 9. Incidence Angle Map
+
+The local incidence angle is defined as the angle between the incident radar signal and the local surface normal.  Output is a geotiff in radians giving the local incidence angle for each pixel in the scene.
+
+*Note that this file is optional. Select the "Incidence angle" option in the HyP3 Processing Options to include it in the product package.*
```
asjohnston-asf, line 137: If we use the same labels as we did for RTC, the option the user will see in Vertex is `Include Inc. Angle Map`

Screenshot from 2021-04-23 08-33-07

asjohnston-asf: And the option in the API/SDK would be `include_inc_map` .  I'm open to revising those if needed, but I think they've served us reasonably well for RTC.

hyp3_gamma/insar/ifm_sentinel.py
```diff
@@ -247,8 +264,8 @@ def make_parameter_file(mydir, parameter_file_name, alooks, rlooks, dem_source):
         f.write('Speckle filtering: off\n')
 
 
-def insar_sentinel_gamma(reference_file, secondary_file, rlooks=20, alooks=4, look_flag=False,
-                         los_flag=False, wrapped_flag=False):
+def insar_sentinel_gamma(reference_file, secondary_file, rlooks=20, alooks=4, include_look_vectors=False,
+                         include_los_deformation=False, include_wrapped_phase=False, include_inc_map=False):
```
asjohnston-asf, line 268: I like the updated parameter names!  I think we use currently `LOS displacement` rather than `LOS deformation` in the README and in Vertex.  I suspect we should use displacement here for consistency?  Or do you prefer we move towards using deformation everywhere?

hyp3_gamma/insar/ifm_sentinel.py
```diff
@@ -233,6 +248,8 @@ def make_parameter_file(mydir, parameter_file_name, alooks, rlooks, dem_source):
         f.write('Baseline: %s\n' % baseline)
         f.write('UTCtime: %s\n' % utctime)
         f.write('Heading: %s\n' % heading)
+        f.write('Spacecraft height %s\n' % height)
+        f.write('Earth radius at nadir %s\n' % erad_nadir)
```
asjohnston-asf, line 252: I think these need colons to be a valid parameter file:
```
        f.write('Spacecraft height: %s\n' % height)
        f.write('Earth radius at nadir: %s\n' % erad_nadir)
```

hyp3_gamma/insar/ifm_sentinel.py
talogan: It should be displacement.  I've changed it in my current copy

PR #257, opened 2021-05-11
undo watermask revert

Jlrine2:


hyp3_gamma/insar/getDemFileGamma.py
```diff
@@ -1,13 +1,18 @@
...
+def get_dem_file_gamma(dem_image: str, dem_par: str, safe_dir: str, pixel_size: int, water_masking: bool):
     geometry = get_geometry_from_kml(f'{safe_dir}/preview/map-overlay.kml')
     with NamedTemporaryFile() as dem_tif:
         prepare_dem_geotiff(dem_tif.name, geometry, pixel_size)
+        if water_masking:
+            apply_water_mask(dem_tif.name, "tmp_masked_file.tif", safe_dir)
+            shutil.move("tmp_masked_file.tif", dem_tif.name)
         execute(f'dem_import {dem_tif.name} {dem_image} {dem_par} - - $DIFF_HOME/scripts/egm2008-5.dem '
-                f'$DIFF_HOME/scripts/egm2008-5.dem_par - - - 1', uselogging=True)
+                f'$DIFF_HOME/scripts/egm2008-5.dem_par - - - 0', uselogging=True)
```
asjohnston-asf, line 18: There's debugging to do here.  Changing the `zflg` parameter to `dem_import` causes values of zero in the source DEM are treated as NoData, causing *all* water (and some valid) land are *always* masked, regardless of the value of the `water_masking` parameter.

We should pick an alternate NoData value for the DEM (potentially impacting both RTC and InSAR), then mask water pixels using that new NoData value when `water_masking` is requested.

hyp3_gamma/dem.py
```diff
@@ -80,11 +82,28 @@ def shift_for_antimeridian(dem_file_paths: List[str], directory: Path) -> List[s
...
+    return:
+        min_val: minimum value of the data type in the geotiff file
+    """
+    ds = gdal.Open(file)
+    dt = gdal_array.GDALTypeCodeToNumericTypeCode(ds.GetRasterBand(1).DataType)
+    v = np.array([1], dtype=dt)
+    if isinstance(v[0], np.floating):
+        return np.finfo(dt).min
+    else:
+        return np.iinfo(dt).min
```
jhkennedy, line 98:
```suggestion
    dtype = GDALTypeCodeToNumericTypeCode(ds.GetRasterBand(1).DataType)
    try:
        np.finfo(dtype).min
    except ValueError:
        np.iinfo(dtype).min
```

hyp3_gamma/dem.py
```diff
@@ -1,11 +1,13 @@
...
+import os
 from pathlib import Path
 from subprocess import PIPE, run
 from tempfile import TemporaryDirectory
 from typing import Generator, List
 
+import numpy as np
 from hyp3lib import DemError
-from osgeo import gdal, ogr
+from osgeo import gdal, gdal_array, ogr
```
jhkennedy, line 10:
```suggestion
from osgeo import gdal, ogr
from osgeo.gdal_array import GDALTypeCodeToNumericTypeCode
```

hyp3_gamma/dem.py
```diff
@@ -110,5 +129,12 @@ def prepare_dem_geotiff(output_name: str, geometry: ogr.Geometry, pixel_size: fl
             gdal.BuildVRT(str(dem_vrt), dem_file_paths)
 
             epsg_code = utm_from_lon_lat(centroid.GetX(), centroid.GetY())
-            gdal.Warp(output_name, str(dem_vrt), dstSRS=f'EPSG:{epsg_code}', xRes=pixel_size, yRes=pixel_size,
+
+            tmp_output = os.path.join(temp_path, 'tmp_output')
+
+            gdal.Warp(tmp_output, str(dem_vrt), dstSRS=f'EPSG:{epsg_code}', xRes=pixel_size, yRes=pixel_size,
```
jhkennedy, line 135:
```suggestion
            tmp_tif = temp_path / 'tmp.tif'
            gdal.Warp(str(tmp_tif), str(dem_vrt), dstSRS=f'EPSG:{epsg_code}', xRes=pixel_size, yRes=pixel_size,
```

hyp3_gamma/dem.py
```diff
@@ -110,5 +129,12 @@ def prepare_dem_geotiff(output_name: str, geometry: ogr.Geometry, pixel_size: fl
...
                       targetAlignedPixels=True, resampleAlg='cubic', multithread=True)
+
+            nodataval = min_value_datatype(dem_file_paths[0])
+
+            gdal.Translate(output_name, tmp_output, noData=nodataval)
```
jhkennedy, line 140:
```suggestion
            gdal.Translate(output_name, str(tmp_tif), noData=nodataval)
```

hyp3_gamma/insar/etc/README_InSAR_GAMMA.txt
```diff
@@ -144,6 +145,15 @@ The digital elevation file contains the terrain height for the scenes being proc
 
 *Note that this file is optional. Select the "Include DEM" option in the HyP3 Processing Options to include it in the product package.*
 
+----------------
+## 11. Water Mask
+
+The water mask specifies, pixel-by-pixel, where large waterbodies exist in the corresponding InSAR product. Pixel values of 0 indicate areas that should be masked out, while areas with a pixel value of 1 are included in the InSAR processing. This mask is applied to the DEM prior to running InSAR jobs, effectively clipping out the water, as areas with DEM heights of 0 are considered invalid for InSAR processing. The water mask is stored as a byte-valued GeoTIFF file and is included in the product package when water masking is applied.
```
jhkennedy, line 151: > as areas with DEM heights of 0 are considered invalid for InSAR processing.

I don't think this is true anymore, right?

hyp3_gamma/insar/getDemFileGamma.py
```diff
@@ -5,9 +5,9 @@
 from hyp3_gamma.dem import get_geometry_from_kml, prepare_dem_geotiff
 
 
-def get_dem_file_gamma(dem_image: str, dem_par: str, safe_dir: str, pixel_size: int):
+def get_dem_file_gamma(dem_image: str, dem_par: str, safe_dir: str, pixel_size: int, water_masking: bool):
```
jhkennedy, line 8: `water_masking` parameter doesn't seem to be used here?

hyp3_gamma/insar/ifm_sentinel.py
```diff
@@ -176,6 +179,11 @@ def move_output_files(output, reference, prod_dir, long_output, include_los_disp
         outName = "{}_inc_map.tif".format(os.path.join(prod_dir, long_output))
         shutil.copy(inName, outName)
 
+    if water_masking:
+        inName = "final_mask.tif"
```
jhkennedy, line 183: I think I'd prefer this to be `final_water_mask.tif` or something with "water" in it.

hyp3_gamma/insar/ifm_sentinel.py
```diff
@@ -176,6 +179,11 @@ def move_output_files(output, reference, prod_dir, long_output, include_los_disp
...
+        outName = "{}_water_mask.tif".format(os.path.join(prod_dir, long_output))
```
jhkennedy, line 184:
```suggestion
        outName = f"{os.path.join(prod_dir, long_output)}_water_mask.tif"
```

hyp3_gamma/insar/ifm_sentinel.py
```diff
@@ -310,15 +318,15 @@ def insar_sentinel_gamma(reference_file, secondary_file, rlooks=20, alooks=4, in
...
     log.info("Got dem of type {}".format(dem_source))
 
     # Figure out which bursts overlap between the two swaths
     burst_tab1, burst_tab2 = get_burst_overlaps(reference_file, secondary_file)
 
     log.info("Finished calculating overlap - in directory {}".format(os.getcwd()))
-    shutil.move(burst_tab1, reference_date_short)
-    shutil.move(burst_tab2, secondary_date_short)
+    shutil.move(burst_tab1, "{}/{}".format(reference_date_short, burst_tab1))
+    shutil.move(burst_tab2, "{}/{}".format(secondary_date_short, burst_tab2))
```
jhkennedy, line 329:
```suggestion
    shutil.move(burst_tab1, f'{reference_date_short}/{burst_tab1}')
    shutil.move(burst_tab2, f'{secondary_date_short}/{burst_tab2}')
```

hyp3_gamma/insar/ifm_sentinel.py
```diff
@@ -367,17 +375,25 @@ def insar_sentinel_gamma(reference_file, secondary_file, rlooks=20, alooks=4, in
     log.info("Starting phase unwrapping and geocoding")
     unwrapping_geocoding(reference, secondary, step="man", rlooks=rlooks, alooks=alooks)
 
+    if water_masking:
+        mask = None
+        tiffiles = glob.glob("./*.tif")
+        for tiffile in tiffiles:
+            if mask is None:
+                mask = get_water_mask(tiffile, reference_file, mask_value=1)
+            apply_water_mask(tiffile, reference_file, mask=mask)
```
jhkennedy, line 384:
```suggestion
    if water_masking:
        tiffiles = glob.glob("./*.tif")
        mask = get_water_mask(tiffiles[0], reference_file, mask_value=1)
        for tiffile in tiffiles:
            apply_water_mask(tiffile, reference_file, mask=mask)
```

sicne your only getting the mask based on the first tif

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -4,18 +4,65 @@
...
+        return np.iinfo(dt).min
+    elif gammadtype in [1]:
+        dt = np.int16
+        return np.iinfo(dt).min
+    elif gammadtype in [2, 3]:
+        dt = np.float32
+        return np.finfo(dt).min
+    else:
+        dt = np.float64
+        return np.finfo(dt).min
```
jhkennedy, line 37:
```suggestion
def get_minimum_value_for_gamma_dtype(dtype):
    """Get the minimum numeric value for a GAMMA data type
    
    GAMMA provides these data types:
    * 0: RASTER 8 or 24 bit uncompressed raster image, SUN (`*.ras`), BMP:(`*.bmp`), or TIFF: (`*.tif`)
    * 1: SHORT integer (2 bytes/value)
    * 2: FLOAT (4 bytes/value)
    * 3: SCOMPLEX (short complex, 4 bytes/value)
    * 4: FCOMPLEX (float complex, 8 bytes/value)
    * 5: BYTE
    """
    dtype_map = {
        0: np.int8,
        1: np.int16,
        2: np.float32,
        3: np.float32,
        4: np.float64,
        5: np.int8,
    }
    try:
        return np.iinfo(dtype_map[dtype]).min
    except ValueError:
        return np.finfo(dtype_map[dtype]).min
    except KeyError:
        raise ValueError(f'Unknown GAMMA data type: {dtype}')

File: hyp3_gamma/insar/unwrapping_geocoding.py [hunk 13]
```diff
@@ -4,18 +4,65 @@
...
+    """
+    ds = gdal.Open(file, gdal.GA_Update)
+    for i in range(1, ds.RasterCount + 1):
+        band = ds.GetRasterBand(i)
+        band_data = band.ReadAsArray()
+        mask = band.GetMaskBand()
+        mask_data = mask.ReadAsArray()
+        band_data[mask_data == 0] = nodata
+        band.WriteArray(band_data)
+        band.SetNoDataValue(float(nodata))
```
jhkennedy, line 54:
```suggestion
    for i in range(1, ds.RasterCount + 1):
        band = ds.GetRasterBand(i)
        band_data = band.ReadAsArray()
        mask_data = band.GetMaskBand().ReadAsArray()
        
        band_data[mask_data == 0] = nodata
        band.WriteArray(band_data)
        band.SetNoDataValue(float(nodata))
```

I'm also unsure why the `float()` around `nodata` is needed -- I'd expect the caller to pass a valid nodata value

hyp3_gamma/insar/water_mask.py
```diff
@@ -0,0 +1,142 @@
...
+            geom = ogr.CreateGeometryFromWkb(transformed.ExportToWkb())
+            defn = outlayer.GetLayerDefn()
+            feat = ogr.Feature(defn)
+            feat.SetField('id', i)
+            feat.SetGeometry(geom)
+            outlayer.CreateFeature(feat)
+            i += 1
+            feat = None
+
+    ds = None
```
jhkennedy, line 51:
```suggestion
    del ds
```

hyp3_gamma/insar/water_mask.py
```diff
@@ -0,0 +1,142 @@
...
+    transform = osr.CoordinateTransformation(sourceprj, targetprj)
+    to_fill = ogr.GetDriverByName("Esri Shapefile")
+    ds = to_fill.CreateDataSource(outshape)
+    outlayer = ds.CreateLayer('', targetprj, ogr.wkbPolygon)
+    outlayer.CreateField(ogr.FieldDefn('id', ogr.OFTInteger))
+
+    # fill the empty layer, converting intersecting features to
+    # a new shapefile
+    i = 0
+    for feature in layer:
```
jhkennedy, line 36:
```suggestion
    for i, feature in enumerate(layer):
```

hyp3_gamma/insar/water_mask.py
```diff
@@ -0,0 +1,142 @@
...
+            transformed = feature.GetGeometryRef()
+            transformed.Transform(transform)
+
+            geom = ogr.CreateGeometryFromWkb(transformed.ExportToWkb())
+            defn = outlayer.GetLayerDefn()
+            feat = ogr.Feature(defn)
+            feat.SetField('id', i)
+            feat.SetGeometry(geom)
+            outlayer.CreateFeature(feat)
+            i += 1
```
jhkennedy, line 48:
```suggestion
```

hyp3_gamma/insar/water_mask.py
```diff
@@ -0,0 +1,142 @@
...
+            feat = None
```
jhkennedy, line 49:
```suggestion
            del feat
```

hyp3_gamma/insar/water_mask.py
```diff
@@ -0,0 +1,142 @@
...
+    dst_rb.Fill(0)
+    dst_rb.SetNoDataValue(0)
+    dst_ds.SetGeoTransform(geotransform)
+
+    _ = gdal.RasterizeLayer(dst_ds, [mask_value], src_lyr)
+    dst_ds.FlushCache()
+    mask = dst_ds.GetRasterBand(1).ReadAsArray()
+    del dst_ds
+    # save the mask file
+    saa.write_gdal_file_byte("final_mask.tif", trans, proj, mask)
```
jhkennedy, line 96:
```suggestion
    saa.write_gdal_file_byte("final_water_mask.tif", trans, proj, mask)
```

hyp3_gamma/insar/water_mask.py
```diff
@@ -0,0 +1,142 @@
...
+    return mask
+
+
+def apply_water_mask(tiffile, safe_dir, mask=None):
+    """
+
+    Given a tiffile input, update the tiffile by filling in all water areas with the maskval.
+
+    """
```
jhkennedy, line 105: Docstring doesn't seem to match

hyp3_gamma/insar/water_mask.py
```diff
@@ -0,0 +1,142 @@
...
+                        default=None)
+    args = parser.parse_args()
+
+    logFile = "apply_water_mask_{}_log.txt".format(os.getpid())
+    logging.basicConfig(filename=logFile, format='%(asctime)s - %(levelname)s - %(message)s',
+                        datefmt='%m/%d/%Y %I:%M:%S %p', level=logging.DEBUG)
+    logging.getLogger().addHandler(logging.StreamHandler())
+    logging.info("Starting run")
+
+    apply_water_mask(args.tiffile, args.outfile, args.safedir, maskval=args.mask_value)
```
jhkennedy, line 142: If we're going to provide this as a script as well, this should be a `main()` function that's registered as an entrypoint in `setup.py`

hyp3_gamma/dem.py
cirrusasf: accept

hyp3_gamma/dem.py
cirrusasf: accepted

hyp3_gamma/dem.py
cirrusasf: changed

hyp3_gamma/dem.py
cirrusasf: changed

cirrusasf: changed

cirrusasf: changed

hyp3_gamma/insar/getDemFileGamma.py
cirrusasf: we do not use water-masking to the dem.tif, instead, we use water-masking in the next right before we output the geotiff file. It turns to be good results. So there is no need to input the water-masking argument.

Keep it as it is.

hyp3_gamma/insar/ifm_sentinel.py
cirrusasf: changed to "final_water_mask.tif", and also made change in water_mask.py

hyp3_gamma/insar/ifm_sentinel.py
cirrusasf: leave it as its to make the statement has the same style as other similar statements. If it is necessary to change, we may also change the other similar statements.
suggest keep it as it is.

hyp3_gamma/insar/ifm_sentinel.py
cirrusasf: changed. smart way!

hyp3_gamma/insar/getDemFileGamma.py
jhkennedy: So you should remove this water masking argument to this call

hyp3_gamma/insar/unwrapping_geocoding.py
cirrusasf: changed it

hyp3_gamma/insar/unwrapping_geocoding.py
cirrusasf: nodata is type of numpy.float32, SetNoDataValue requires the float type. so We need float() there. When I debug the progam, I remember one time mask_data = band.GetMaskBand().ReadAsArray() does not work, this is why I use two statements instead of one.

jhkennedy: Ah okay, so does the cast to float work with integer types? 

This pattern
```python
nodata = get_minimum_value_gamma_dtype(type_)
setnodata(outname, nodata)
```
can return int or float types, so I would suspect `setnodata` with `float(dtype)` wouldn't work for integer data types. Or does the conversion to tif from GAMMA format always create float tifs? If so, the `get_minimum_value_gamma_dtype` could just convert everything to `float` before it returns.

cirrusasf: I tested that float() works for int. I didn't check if GAMMA format always create float tifs.

hyp3_gamma/insar/getDemFileGamma.py
cirrusasf: Yes. I delete the water_masking. Sorry to the confusing.

hyp3_gamma/insar/unwrapping_geocoding.py
jhkennedy: :+1: I think I'd move the `float()` into `get_minimum_value_gamma_dtype` so that `setnodata` doesn't hide that magic

jhkennedy: Actually scratch that, it's fine -- GDAL apparently typically assumes the nodata value is a `float`. E.g., [`gdal_edit.py`](https://github.com/OSGeo/gdal/blob/master/gdal/swig/python/gdal-utils/osgeo_utils/gdal_edit.py#L135)

cirrusasf:
Hi Joe,

Do you mean we don't drop the float() ?

Jiang

hyp3_gamma/insar/unwrapping_geocoding.py
cirrusasf: do you mean we leave float() there?

jhkennedy: Yeah, leave the float there -- seems to be a GDAL thing. :-D

cirrusasf: OK.

hyp3_gamma/insar/getDemFileGamma.py
cirrusasf: new code inserts the nodata=minimum_valule_datatype into the dem.tif, then use this dme.tif in the gamma process. In the dem_import, we use execute(f'dem_import {dem_tif.name} {dem_image} {dem_par} - - $DIFF_HOME/scripts/egm2008-5.dem '  f'$DIFF_HOME/scripts/egm2008-5.dem_par - - - - -', uselogging=True).

hyp3_gamma/insar/etc/README_InSAR_GAMMA.txt
cirrusasf: I changed the words to reflect how the water mask is applied to the proecess.

CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [4.7.0](https://github.com/ASFHyP3/hyp3-gamma/compare/v4.6.0...v4.7.0)
+
+### Added
+* Option to apply water masking to output geotiff files produced by gamma processing
+  * define the nodata value (minimum_value_datatype) to the dem.tif
+    and apply this dem.tif to gamma processing
```
jhkennedy, line 14:
```suggestion
* Option to apply a water mask to DEM before processing and include the water mask GeoTIFF with InSAR products
```

jhkennedy (changes requested):
~~One minor comment on the changelog, and~~ you should remove the `hyp3_gamma/insar/etc/README_InSAR_GAMMA.txt` file from this PR, but otherwise looks good!

Note: README language will go in hyp3-metadata-templates (via ASFHyP3/hyp3-metadata-templates#97) and before release, we'll need to do _another_ PR for updating the metadata templates.

hyp3_gamma/dem.py
```diff
@@ -110,5 +128,10 @@ def prepare_dem_geotiff(output_name: str, geometry: ogr.Geometry, pixel_size: fl
...
 
             epsg_code = utm_from_lon_lat(centroid.GetX(), centroid.GetY())
-            gdal.Warp(output_name, str(dem_vrt), dstSRS=f'EPSG:{epsg_code}', xRes=pixel_size, yRes=pixel_size,
+
+            tmp_tif = temp_path / 'tmp.tif'
+            gdal.Warp(str(tmp_tif), str(dem_vrt), dstSRS=f'EPSG:{epsg_code}', xRes=pixel_size, yRes=pixel_size,
                       targetAlignedPixels=True, resampleAlg='cubic', multithread=True)
+
+            nodataval = min_value_datatype(dem_file_paths[0])
+            gdal.Translate(output_name, str(tmp_tif), noData=nodataval)
```
asjohnston-asf, line 137: There may not be a need for the separate `gdal.Translate` call to assign the new NoData value, I think we might get the same results by simply setting the `dstNodata=nodataval` option in the `gdal.Warp` call

hyp3_gamma/insar/getDemFileGamma.py
```diff
@@ -10,4 +10,4 @@ def get_dem_file_gamma(dem_image: str, dem_par: str, safe_dir: str, pixel_size:
     with NamedTemporaryFile() as dem_tif:
         prepare_dem_geotiff(dem_tif.name, geometry, pixel_size)
         execute(f'dem_import {dem_tif.name} {dem_image} {dem_par} - - $DIFF_HOME/scripts/egm2008-5.dem '
-                f'$DIFF_HOME/scripts/egm2008-5.dem_par - - - 1', uselogging=True)
+                f'$DIFF_HOME/scripts/egm2008-5.dem_par - - - - -', uselogging=True)
```
asjohnston-asf, line 13: `-` just means "use the default value", there's no need to provide trailing `-` parameters if we're not actively setting any later parameters
```suggestion
                f'$DIFF_HOME/scripts/egm2008-5.dem_par', uselogging=True)
```

CHANGELOG.md
jhkennedy: okay, my mistake, the changlog as stated is more accurate.

hyp3_gamma/dem.py
cirrusasf: I tested gdal.warp with option dstNodata=nodataval, it does not work.  This is why I use separate  gdal.Translate to do it.

hyp3_gamma/insar/getDemFileGamma.py
cirrusasf: The call of the dem_import is:
dem_import <input_DEM> <DEM> <DEM_par> [input_type] [priority] [geoid] [geoid_par] [geoid_type] [latN_shift] [lonE_shift] [zflg] [no_data]
I didn't check if we do not provide [no_data] option, what the result will be. I suggest keep the [no_data] option, so that we get clear understanding which option is which when we compare with the call format of the command dem_import.

CHANGELOG.md
```diff
@@ -29,6 +39,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Changed
 * `conda-env.yml` has been renamed to `environment.yml` in-line with community practice
+
```
Jlrine2, line 42:
```suggestion
```

Jlrine2:
LGTM, but I cannot approve since I am the Author, either @asjohnston-asf @jhkennedy should also take a look

cirrusasf (approved):

jhkennedy:
TruffleHog failure is due to a dirty commit with unresolved merge conflicts in the commit history, but not in any current files. Fine to ignore as it'll go away next release.

jhkennedy (approved):

PR #271, opened 2021-06-15
remove safe off of the reference granule in metadata

Jlrine2:


hyp3_gamma/insar/ifm_sentinel.py
```diff
@@ -292,7 +292,8 @@ def insar_sentinel_gamma(reference_file, secondary_file, rlooks=20, alooks=4, in
     secondary_date = secondary_file[17:32]
     secondary_date_short = secondary_file[17:25]
     igramName = "{}_{}".format(reference_date, secondary_date)
-
+    reference_granule = os.path.splitext(os.path.basename(reference_file))[0]
+    secondary_granule = os.path.splitext(os.path.basename(secondary_file))[0]
```
asjohnston-asf, line 296: Should define these immediately before they're used, unless we're going to refactor to leverage them other function calls

asjohnston-asf (approved):

PR #293, opened 2021-07-15
Water mask refactor

asjohnston-asf:


hyp3_gamma/water_mask.py
```diff
@@ -23,6 +23,7 @@ def create_water_mask(input_tif: str, output_tif: str):
     dst_ds = gdal.GetDriverByName('GTiff').Create(output_tif, src_ds.RasterXSize, src_ds.RasterYSize, 1, gdal.GDT_Byte)
     dst_ds.SetGeoTransform(src_ds.GetGeoTransform())
     dst_ds.SetProjection(src_ds.GetProjection())
+    dst_ds.SetMetadata(src_ds.GetMetadata())
```
asjohnston-asf, line 26: Setting *all* the metadata based on the input tif seems unecessary. Can we just set the `AREA_OR_POINT` field based on the source value?

asjohnston-asf: This might work?
```
dst_ds.SetMetadataItem('AREA_OR_POINT', src_ds.GetMetadataItem('AREA_OR_POINT')
```

cirrusasf: Yes. We can.

cirrusasf: I will change this and test it.

cirrusasf (approved):

PR #294, opened 2021-07-15
Finishing touches for water mask tests

asjohnston-asf:


tests/test_water_mask.py
```diff
@@ -16,3 +17,32 @@ def test_create_water_mask(tmp_path, test_data_dir):
     assert info['bands'][0]['type'] == 'Byte'
     assert info['bands'][0]['minimum'] == 1
```
cirrusasf, line 18: should it be assert info['bands'][0]['minimum'] == 0 ?

asjohnston-asf: This particular geotiff is inland, so I'm asserting that all the pixels have a value of 1.

cirrusasf: OK. I got it. forget about it.

cirrusasf (approved):
it looks to me.

PR #304, opened 2021-08-10
Water mask applied metadata

jhkennedy:
Will pull in the the upcoming 1.3.0 release of `hyp3_datadata` with the water mask applied description (ASFHyP3/hyp3-metadata-templates#110).


jhkennedy:
TruffleHog failure is due to a dirty commit with unresolved merge conflicts in the commit history, but not in any current files. Fine to ignore as it'll go away next release.

cirrusasf (approved):

PR #303, opened 2021-08-06
upgrade to GAMMA 20210701

asjohnston-asf:
Major version upgrades:
- Ubuntu `20.04`, instead of `18.04`
- GAMMA `20210701`, instead of a patched version of `20191203`
- system Python `3.8.10`, instead of `3.6.9`
- GDAL `3.0.4`, instead of `2.2.3`

PR #310, opened 2021-08-29
wrap unw_phase.png at 12*pi

asjohnston-asf:
The `rasdt_pwr` call to create the `unw_phase.png` browse image was incorrectly implemented in the [initial gamma upgrade PR](https://github.com/ASFHyP3/hyp3-gamma/pull/303):
```
< rasrmg {ifgname}.adf.unw {mmli} {width} 1 1 0 1 1 0.33333 1.0 .35 0.0 - {ifgname}.adf.unw.ras
> rasdt_pwr {ifgname}.adf.unw {mmli} {width} - - - - - - 1 rmg.cm {ifgname}.adf.unw.ras
```
The fringes were incorrectly wrapped to mod `1.0` instead of mod `12*pi`. This fixes the wrapping. The fixed fringes start at a different value compared with production, but I'm comfortable leaving that unaddressed.

Production:
prod

Test:
test

Fixed:
fixed

### DIFF
--- a/hyp3_gamma/insar/unwrapping_geocoding.py
+++ b/hyp3_gamma/insar/unwrapping_geocoding.py
@@ -2,7 +2,6 @@
 
 import argparse
 import logging
-import math
 import os
 from tempfile import TemporaryDirectory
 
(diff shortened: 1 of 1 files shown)


jhkennedy (approved):
Just, one super minor thing that doesn't really matter

PR #315, opened 2021-09-16
output the reference point coordinates

cirrusasf:


hyp3_gamma/insar/ifm_sentinel.py
```diff
@@ -286,6 +286,14 @@ def make_parameter_file(mydir, parameter_file_name, alooks, rlooks, dem_source):
...
         f.write('DEM resolution (m): %s\n' % (res * 2))
         f.write('Unwrapping type: mcf\n')
+        f.write('Reference Point_sar_azlin: %s\n' % coords["row_s"])
+        f.write('Reference Point_sar_rpix: %s\n' % coords["col_s"])
+        f.write('Reference Point_map_row: %s\n' % coords["row_m"])
+        f.write('Reference Point_map_col: %s\n' % coords["col_m"])
+        f.write('Reference Point_map_y: %s\n' % coords["y"])
+        f.write('Reference Point_map_x: %s\n' % coords["x"])
+        f.write('Reference Point_map_lat: %s\n' % coords["lat"])
+        f.write('Reference Point_map_lon: %s\n' % coords["lon"])
```
jhkennedy, line 296: Are `Point_sar_azli` and similar meaningful names? The of the lines in this parameter file use plain language names.

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -16,6 +17,45 @@
...
+        coord_lst = ' '.join(coord_str[0].split()).split(" ")[: -1]
+        coord_lst = [float(s) for s in coord_lst]
+        return coord_lst
+
+    cmd = ['sarpix_coord', in_mli_par, '-']
+    coords = {}
+    if in_dem_par:
+        cmd1 = cmd.copy()
+        cmd1.extend([in_dem_par])
+        cmd1.extend([str(ref_azlin), str(ref_rpix)])
```
jhkennedy, line 38: In general, when sub process commands are made, it's much better to explicitly list the whole command, even if it means repetition

And f-string is easiest to read, but an explicit list is okay too. 

(same for next two places)

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -127,16 +168,21 @@ def unwrapping_geocoding(reference, secondary, step="man", rlooks=10, alooks=2,
 
     execute(f"rascc_mask {ifgname}.adf.cc {mmli} {width} 1 1 0 1 1 0.10 0.20 ", uselogging=True)
 
+    # default reference point in SAR space
+    ref_azlin, ref_rpix = 0, 0
+
+    coords = get_coords(f"{mmli}.par", ref_azlin, ref_rpix, dempar)
```
jhkennedy, line 174: This can just be default values for `get_coords` instead of hard coding them here, can then you can call it like

```python
    coords = get_coords(f"{mmli}.par", in_dem_par=dempar)
```

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -16,6 +17,45 @@
...
+    inputs: mil.par, dempar, reference point  in SAR space (ref_azlin, ref_rpix)
+    returns: coords={"row_s":row_s,"col_s":col_s,"row_m":row_m,"col_m":col_m,"x":x,"y":y,"lat":lat,"lon":lon}
+    """
+    def _coord_lst(cmd):
+        coord_txt = subprocess.run(cmd, capture_output=True, text=True)
+        lst = coord_txt.stdout.split('\n')
+        coord_str = [s for s in lst if "selected" in s]
+        coord_lst = ' '.join(coord_str[0].split()).split(" ")[: -1]
+        coord_lst = [float(s) for s in coord_lst]
+        return coord_lst
```
jhkennedy, line 31: I think it'd be better to move this function outside of `get_coords` and give it a more descriptive name. 

Really, this is `coords_from_sarpix_coord`, and could have the command template inside where you provide the command arguments as parameters

hyp3_gamma/insar/unwrapping_geocoding.py
cirrusasf: changed

hyp3_gamma/metadata/templates/insar/readme.md.txt.j2
```diff
@@ -98,11 +98,11 @@ The wrapped interferogram image is generated from the interferogram phase prior
...
 
-Phase unwrapping is most effective when the reference point is beyond the area of deformation and there are not any areas with deformation greater than half the sensor wavelength within the space of a single fringe. Because choosing an optimal reference point requires knowledge of the site characteristics and examination of the interferogram, it is not practical in an automated, global workflow. By default, the first point of the combined scene is used as the reference point for ASF's On Demand InSAR products.
+Phase unwrapping is most effective when the reference point is beyond the area of deformation and there are not any areas with deformation greater than half the sensor wavelength within the space of a single fringe. Because choosing an optimal reference point requires knowledge of the site characteristics and examination of the interferogram, it is not practical in an automated, global workflow. By default, ASF's On Demand InSAR products use the location of the pixel with the highest coherence value as the reference point, and the phase difference at this location is set to 0 during phase unwrapping.
```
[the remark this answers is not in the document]
hjkristenson: Ha, just re-read the JIRA description, and we listed both the coordinates and the original phase value. Whoops. I forgot about that when I was working with @jhkennedy to get the coordinates set up as a variable.

jhkennedy (approved):

PR #330, opened 2021-11-09
Pixelshift

cirrusasf:

### DIFF
--- a/hyp3_gamma/insar/unwrapping_geocoding.py
+++ b/hyp3_gamma/insar/unwrapping_geocoding.py
@@ -4,6 +4,7 @@ import argparse
 import logging
 import os
 import subprocess
+import glob
 from tempfile import TemporaryDirectory
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -109,7 +111,7 @@ def geocode(inname, outname, inwidth, lt, outwidth, outlines, type_):
     execute(f"geocode {lt} {inname} {inwidth} {outname} {outwidth} {outlines} - {type_}", uselogging=True)
 
 
-def data2geotiff(inname, outname, dempar, type_):
+def data2geotiff(inname, outname, dempar, type_, shift_origin=False):
```
asjohnston-asf, line 114: `shift_origin` should be removed if it isn't being used
```suggestion
def data2geotiff(inname, outname, dempar, type_):
```

jhkennedy:
@cirrusasf `bumpless` isn't an appropriate label to be attaching to these PRs.

jhkennedy:
@cirrusasf I believe you've determined we don't need to do the shift for COP30 and we don't want to change the legacy products (would rather retire if we're not being consistent), is that correct? 

I think we can close this PR if so.

cirrusasf:
> I checked the both RTC and INSAR product with Copernicus DEM. There is no pixel shift issue based on comparison of the product against the corner reflectors. We could close the PR now. I would rather wait till next week when Tom came back.

PR #331, opened 2021-12-01
Area point

cirrusasf:


cirrusasf:
Tom, Joe, and I discussed the half pixel issue before. We would like to leave this issue along, since the half-pixel issue only occurs when we use legacy DEM. So I close this PR.

PR #335, opened 2021-12-11
Cc pwr thres

cirrusasf:


hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -85,6 +85,28 @@ def read_bmp(file):
...
+    out_im.save(out_file)
+    return out_file
+
+
+def convert_bmp_bin(bmpfile: str, outfile: str):
+    data = read_bmp(bmpfile)
+    data.tofile(outfile)
+    return outfile
+
+
```
jhkennedy, line 109: These don't seem to be used
```suggestion
```

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -101,6 +123,11 @@ def ref_point_with_max_cc(fcc: str, fmask: str, mlines: int, mwidth: int):
     return 0, 0
 
 
+def get_avg_intensity(mmli: str, mlines: int, mwidth: int):
+    data = read_bin(mmli, mlines, mwidth)
+    return float(data.mean())
+
+
```
jhkennedy, line 130: This doesn't seem to be used
```suggestion
```

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -117,64 +144,72 @@ def create_phase_from_complex(incpx, outfloat, width):
...
 
-def convert_from_sar_2_map(in_file, out_geotiff, width, lt, dempar, demw, demn, type_):
+def convert_from_sar_2_map(in_file, out_geotiff, mwidth, lt, dempar, demw, demn):
+    """
+    in_file is a float binary file, and out_geotiff is a geotiff file.
+    """
+    geocode_back(in_file, "tmp", mwidth, lt, demw, demn, 0)
+    data2geotiff("tmp", out_geotiff, dempar, 2)
+
+
```
jhkennedy, line 174: These don't seem to be used
```suggestion
```

jhkennedy (changes requested):
I really can't tell what's all going on in this PR. From what I can tell, I think the only change we're trying to make is to this command:
```
execute(f"rascc_mask {ifgname}.adf.cc {mmli} {width} 1 1 0 1 1 0.10 0.20 ", uselogging=True)
```
and I was expecting `0.10` and `0.20` to be reversed.

It looks like there's a pile of unrelated changes from an point in time where the `--apply-water-mask` option was being worked on. I tried merging in `develop`, but it still contains changes related to applying the water mask -- possibly even walking back some of the desired changes there. 

I think, it'd be better to close this PR and start a new feature branch from develop, with *just* the changes related to the cc and pwr threshold

cirrusasf:
close the pull request. Start the new branch cc_pwr_thres_v2. work over there.

PR #337, opened 2022-02-01
Earthdata creds

jtherrmann:
It is now possible to inject Earthdata username and password as environment variables, though cli args take precedence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 5.1.3
+
+### Security
+
+* It is now possible to inject Earthdata username and password using environment variables.
 
 ## [5.1.2](https://github.com/ASFHyP3/hyp3-gamma/compare/v5.1.1...v5.1.2)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### Security
 
-* It is now possible to inject Earthdata username and password using environment variables.
+* It is now possible to inject Earthdata username and password using environment variables: `EARTHDATA_USERNAME`
+  and `EARTHDATA_PASSWORD`.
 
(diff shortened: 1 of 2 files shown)


hyp3_gamma/__main__.py
```diff
@@ -33,10 +34,20 @@ def main():
     )
 
 
+def write_earthdata_creds(username, password):
+    if username is None:
+        username = os.environ['EARTHDATA_USERNAME']
```
asjohnston-asf, line 39: I believe this will throw a `KeyError` if the environment variable is not defined? Should we give the user an explicit error message, i.e. "earthdata login username must be provided via either the --username option or an EARTHDATA_USERNAME environment variable"?

hyp3_gamma/__main__.py
```diff
@@ -33,10 +34,20 @@ def main():
     )
 
 
+def write_earthdata_creds(username, password):
```
asjohnston-asf, line 37: I slightly prefer to avoid abbreviations, so I'd suggest `credentials` over `creds`, but that's not a sticking point

hyp3_gamma/__main__.py
jtherrmann: Good point. Would it be best to write that message to stderr?

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 5.1.3
+
+### Security
+
+* It is now possible to inject Earthdata username and password using environment variables.
```
asjohnston-asf, line 13: I think it's worth calling out the names of the environment variables here, so a casual user won't have to dig through the implementation to find them, but again it's not a sticking point since I'm not sure any such users exist at the moment

hyp3_gamma/__main__.py
asjohnston-asf: We typically `raise` a python exception with a custom error message, e.g. 
https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/rtc/rtc_sentinel.py#L91
https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/rtc/rtc_sentinel.py#L138
https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/rtc/rtc_sentinel.py#L180

asjohnston-asf (approved):

jhkennedy (approved):

PR #339, opened 2022-02-03
pin pandas version in python dependencies

asjohnston-asf:

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,6 +19,7 @@ dependencies:
   # For running
   - gdal
   - geopandas
+  - pandas>=1.3.5,<1.4
   - hyp3lib>=1.6.8,<2
   - importlib_metadata
   - jinja2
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -19,6 +19,7 @@ dependencies:
   # For running
   - gdal
   - geopandas
+  - pandas>=1.3.5,<1.4
```
jhkennedy, line 22: I think this could be simplified with `pandas=1.3`, but they are effectively the same and this is more explicit, so either is fine by me.

jhkennedy (approved):
This should be fin

environment.yml
jhkennedy: (and would need to be `pandas==1.3` in `setup.py`)

jhkennedy:
Closing due to this being fixed in #340

PR #340, opened 2022-02-03
upgrade pip and pin numpy version to match system ubuntu version

asjohnston-asf:
The apt install puts v1.17.4 in the container. With this approach, the subsequent pip install doesn't change numpy and correctly brings in pandas 1.3.5.

Numpy 1.17.4 isn't available via conda-forge, but 1.17.5 is, hence not pinning either setup.py or environment.yml to a specific patch version.


jhkennedy (approved):
I agree, I like this better.

PR #341, opened 2022-02-04
Encrypt Earthdata Login Information 

jacquelynsmale:
### Developer checklist

- [x ] Assigned a reviewer
- [x ] Indicated the level of changes to this package by affixing one of these labels:
  * major -- Major changes to the API that may break current workflows
  * minor -- Minor changes to the API that do not break current workflows
  * patch -- Patches and bugfixes for the current version that do not break current workflows
  * bumpless -- Changes to documentation, CI/CD pipelines, etc. that don't affect the software's version
- [ ] (If applicable) Added/updated documentation for these changes
- [ ] (If applicable) Added/updated tests for these changes


.github/workflows/test-and-build.yml
```diff
@@ -17,42 +17,30 @@ env:
...
 
   dockerize:
+    needs: call-version-info-workflow
     runs-on: ubuntu-latest
+    env:
+      SDIST_VERSION: ${{ needs.call-version-info-workflow.outputs.version_tag }}
     steps:
       - uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
```
jhkennedy, line 42: I think we'll need to add this back in like in ASFHyP3/actions#8

.github/workflows/test-and-build.yml
```diff
@@ -17,42 +17,30 @@ env:
...
-      - uses: conda-incubator/setup-miniconda@v2
-        with:
-          mamba-version: "*"
-          activate-environment: hyp3-gamma
-          environment-file: environment.yml
 
       - name: set environment variables
         shell: bash -l {0}
         run: |
-          git fetch origin +refs/tags/*:refs/tags/*
```
jhkennedy, line 53: I think we'll need to add this back in like ASFHyP3/actions#8

jhkennedy (changes requested):
I'll open a PR for these changes -- will confirm they are needed via an RTC or InSAR product README

jhkennedy (approved):
Confirmed #342 fixed the package version number; this should be g2g!

PR #342, opened 2022-02-04
fetch tags for in-docker package build

jhkennedy:
Duplicating ASFHyP3/actions#8 here since we can't use the re-usable workflows due to GAMMA shenanigans.

### DIFF
--- a/.github/workflows/test-and-build.yml
+++ b/.github/workflows/test-and-build.yml
@@ -37,10 +37,12 @@ jobs:
       SDIST_VERSION: ${{ needs.call-version-info-workflow.outputs.version_tag }}
     steps:
       - uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-and-build.yml
```diff
@@ -37,10 +37,12 @@ jobs:
       SDIST_VERSION: ${{ needs.call-version-info-workflow.outputs.version_tag }}
     steps:
       - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
       - name: set environment variables
-        shell: bash -l {0}
```
jhkennedy, line 42: relic of miniconda setup previously -- does no harm but is no longer needed.

asjohnston-asf (approved):

PR #347, opened 2022-02-18
test getting rid of power threshold when calcualting vadility mask

cirrusasf:


jhkennedy:
Looks good to me -- I've not approved it because we should get the COP GeoJSON out the door first. When this is ready to be merged, I'll hit the approve button

PR #353, opened 2022-03-02
Rework how we bring in the asf-tools conda env

jhkennedy:
This is an *extremely* short term fix to pin the asf-tools environment used for water mapping back to the last known working environment and also keep `hpy3-gamma` up to date. 

It's directly copying the asf-tools env, so there's no dependency updates or changes and it should function *exactly* like in v5.1.2

This also paves the way for a medium term, or possibly long term, way to manage and develop these environments separately:
* In ASFHyP3/asf-tools#94 we're developing an asf-tools container which can become the source of truth for the asf-tools environment here and this container would always copy in the `latest` asf-tools env out of that container
  * asf-tools development can be managed entirely in the asf-tools space
  * hyp3-gamma develop can continue without worrying about the water mapping environment

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,9 @@
+## ASF TOOLS
+ARG ASF_TOOLS_IMAGE=845172464411.dkr.ecr.us-west-2.amazonaws.com/hyp3-gamma
+ARG ASF_TOOLS_TAG=5.1.2
+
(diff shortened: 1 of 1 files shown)


jhkennedy:
I confirmed this container works for a GRD scene as well as one of the current failures in the `hyp3-watermap` deployment. So when we release this, we can unpin the `hyp3-gamma` version in the watermap deployment

asjohnston-asf (approved):

PR #363, opened 2022-03-30
Add more detail to changelog entry for InSAR updates

hjkristenson:


jhkennedy (approved):
I like the verbose here

PR #364, opened 2022-03-31
fix the error arguments azlin and rpix in the sarpix_coord command

cirrusasf:
This PR solves the issue about the reference point indexes. In previous version, we get fooled by the output of the mcf command. It always reports the reference point as the azlin+1, rpix+1, where azlin and rpix are the arguments of the mcf, and are reference point indexes starting from 0. After testing, we found that the sarpix_coord command accepts the reference point (azlin, rpix) which is based on indexes starting from 0. So we correct the previous code, and output the reference point  indexes stating from 0 in the meta text. We need explain the reference point in the InSAR Product Guide and/or README file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [5.2.1]
+
+### Changed
+* Output the reference point in SAR space in the meta text file. The reference point indexes start form 0.
 ## [5.2.0]
 
 ### Changed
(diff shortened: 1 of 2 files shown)


hjkristenson:
The readme file doesn't go into specifics on any of the other parameters included. Perhaps it would make sense to document each of the lines included in the parameter file in the InSAR Product Guide, and just add a link to the readme template (hyp3_gamma/metadata/templates/insar/readme.md.txt.j2) that points to that section in the product guide. 

The documentation update would take some time, but I don't think it's necessarily required in order to release this code. We haven't offered any explanation of the parameters to this point, so it's not degrading the product. 

We could either update the readme template as part of this PR (and add the requisite section in the InSAR Product Guide), or we can release the code changes and add a story to the backlog to enhance the documentation with the parameter file content descriptions.

jtherrmann:
With regard to the following:
> output the reference point indexes stating from 0 in the meta text
does this just refer to the logging in [this function](https://github.com/ASFHyP3/hyp3-gamma/blob/6910b41db4842ef73f20099b635e4d902d54958d/hyp3_gamma/insar/unwrapping_geocoding.py#L33) or is there somewhere else in the code that logs the reference point?

jtherrmann (approved):

jtherrmann (approved):

PR #365, opened 2022-04-07
Add flood maps

jacquelynsmale:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [5.2.1](https://github.com/ASFHyP3/hyp3-gamma/compare/v4.5.1...v4.6.0)
 
 ### Added
-* `flood_map` entrypoint to create a flood depth map product. 
+* Option to add a `--include-flood-depth` flag when running `water_map` to produce a map of estimated flood depths.
 
 ## [5.2.0]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [5.3.0](https://github.com/ASFHyP3/hyp3-gamma/compare/v4.5.1...v4.6.0)
+## [5.3.0]
 
 ### Added
 * Added `--include-flood-depth` option for `water_map` to produce a map of estimated flood depths.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Added
 * Added `--include-flood-depth` option for `water_map` to produce a map of estimated flood depths.
 
+## [5.2.1]
+
 ### Changed
 * Output the reference point in SAR space in the meta text file. The reference point indexes start form 0.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [5.2.1](https://github.com/ASFHyP3/hyp3-gamma/compare/v4.5.1...v4.6.0)
+
+### Added
+* `flood_map` entrypoint to create a flood depth map product. 
+
 ## [5.2.0]
 
 ### Changed
--- a/hyp3_gamma/__main__.py
+++ b/hyp3_gamma/__main__.py
@@ -187,7 +187,7 @@ def water_map():
 
     parser.add_argument('--include-flood-depth', type=string_is_true, default=False)
     parser.add_argument('--estimator', type=str, default='iterative', choices=['iterative', 'logstat', 'nmad', 'numpy'])
-    parser.add_argument('--water-level-sigma', type=float, default=3., )
+    parser.add_argument('--water-level-sigma', type=float, default=3.)
     parser.add_argument('--known-water-threshold', type=float, default=30.)
     parser.add_argument('--iterative-bounds', type=int, nargs=2, default=[0, 15])
 
--- a/hyp3_gamma/__main__.py
+++ b/hyp3_gamma/__main__.py
@@ -175,63 +175,6 @@ def water_map():
     parser.add_argument('--password')
     parser.add_argument('--bucket')
     parser.add_argument('--bucket-prefix', default='')
-    parser.add_argument('--resolution', type=float, choices=[10.0, 30.0], default=30.0)
-    parser.add_argument('--speckle-filter', type=string_is_true, default=False)
-    parser.add_argument('--max-vv-threshold', type=float, default=-17.)
-    parser.add_argument('--max-vh-threshold', type=float, default=-24.)
-    parser.add_argument('--hand-threshold', type=float, default=15.)
-    parser.add_argument('--hand-fraction', type=float, default=0.8)
-    parser.add_argument('--membership-threshold', type=float, default=0.45)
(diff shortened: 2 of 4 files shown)


hyp3_gamma/__main__.py
```diff
@@ -226,5 +226,75 @@ def water_map():
             upload_file_to_s3(product_file, args.bucket, args.bucket_prefix)
 
 
+def flood_map():
```
jhkennedy, line 229: This looks pretty good! I think we only need one "water" job, however, so you could combine this with the water map job. Essentially, just adding the second conda execute and extra args up to the water map job.

I think we had talked with Franz about it being an extra layer (you can confirm with him), so you'd also need an `--include-flood-depth` true/false option that would toggle running the second execute

jacquelynsmale: Great feedback! I deleted `flood_map()` and added the execution after `if args.include_flood_map:` in `water_map()`.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [5.2.1](https://github.com/ASFHyP3/hyp3-gamma/compare/v4.5.1...v4.6.0)
```
jtherrmann, line 9: Feels like a minor change.
```suggestion
## [5.3.0](https://github.com/ASFHyP3/hyp3-gamma/compare/v4.5.1...v4.6.0)
```

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+
+### Added
+* Option to add a `--include-flood-depth` flag when running `water_map` to produce a map of estimated flood depths.
```
jtherrmann, line 12:
```suggestion
* Added `--include-flood-depth` option for `water_map` to produce a map of estimated flood depths.
```

hyp3_gamma/__main__.py
```diff
@@ -183,6 +184,13 @@ def water_map():
     parser.add_argument('--hand-fraction', type=float, default=0.8)
     parser.add_argument('--membership-threshold', type=float, default=0.45)
     parser.add_argument('granule')
+
```
jtherrmann, line 187: Is it redundant to include default values since they already exist in the job spec? Maybe @asjohnston-asf can weigh in.

hyp3_gamma/__main__.py
```diff
@@ -213,6 +221,13 @@ def water_map():
             f'--hand-threshold {args.hand_threshold} --hand-fraction {args.hand_fraction} '
             f'--membership-threshold {args.membership_threshold}', uselogging=True)
 
+    if args.include_flood_depth:
+        execute(f'conda run -n  asf-tools flood_map {product_name}/{product_name}_FM.tif '
```
jtherrmann, line 225: This command needs to be updated since `flood_map` is no longer a separate command.

hyp3_gamma/__main__.py
```diff
@@ -183,6 +184,13 @@ def water_map():
     parser.add_argument('--hand-fraction', type=float, default=0.8)
     parser.add_argument('--membership-threshold', type=float, default=0.45)
     parser.add_argument('granule')
+
+    parser.add_argument('--include-flood-depth', type=string_is_true, default=False)
+    parser.add_argument('--estimator', type=str, default='nmad', choices=['iterative', 'logstat', 'nmad', 'numpy'])
+    parser.add_argument('--water-level-sigma', type=float, default=3., )
+    parser.add_argument('--known-water-threshold', type=float, default=30.)
+    parser.add_argument('--iterative-bounds', type=float, default=[0, 15])
```
jtherrmann, line 192: Maybe the type should be `tuple` and the default value should be a tuple instead of a list?

hyp3_gamma/__main__.py
asjohnston-asf: I think default values here are preferred, since it's nice to have reasonable examples/default when running the python code or docker container outside of HyP3. I agree it's unfortunate that the defaults have to be written down in two different places, though. We went ahead and lived with the duplication for RTC_GAMMA, for example:
https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/__main__.py#L57
https://github.com/ASFHyP3/hyp3/blob/develop/job_spec/RTC_GAMMA.yml#L30

jtherrmann: Gotcha, that's what I figured.

hyp3_gamma/__main__.py
jtherrmann: Nope, I'm dumb.

hyp3_gamma/__main__.py
```diff
@@ -213,6 +221,13 @@ def water_map():
             f'--hand-threshold {args.hand_threshold} --hand-fraction {args.hand_fraction} '
             f'--membership-threshold {args.membership_threshold}', uselogging=True)
 
+    if args.include_flood_depth:
+        execute(f'conda run -n  asf-tools flood_map {product_name}/{product_name}_FM.tif '
+                f'{product_name}/{product_name}_WM.tif {product_name}/{product_name}_WM_HAND.tif '
+                f'--estimator {args.estimator} --water_level_sigma {args.water_level_sigma} '
+                f'--known_water_threshold {args.known_water_threshold} --iterative-bounds {args.iterative_bounds} ',
```
asjohnston-asf, line 228: now that `args.iterative_bounds` is a list, I think we need to do a little manual formatting here. As written, we'd get `--iterative-bounds [0, 15]`, where I think we need `--iterative-bounds 0 15`

jtherrmann (dismissed):

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [5.3.0](https://github.com/ASFHyP3/hyp3-gamma/compare/v4.5.1...v4.6.0)
```
asjohnston-asf, line 9:
```suggestion
## [5.3.0]
```

hyp3_gamma/__main__.py
```diff
@@ -183,6 +184,13 @@ def water_map():
     parser.add_argument('--hand-fraction', type=float, default=0.8)
     parser.add_argument('--membership-threshold', type=float, default=0.45)
     parser.add_argument('granule')
+
+    parser.add_argument('--include-flood-depth', type=string_is_true, default=False)
+    parser.add_argument('--estimator', type=str, default='nmad', choices=['iterative', 'logstat', 'nmad', 'numpy'])
+    parser.add_argument('--water-level-sigma', type=float, default=3., )
+    parser.add_argument('--known-water-threshold', type=float, default=30.)
+    parser.add_argument('--iterative-bounds', type=int, nargs=2, default=[0, 15])
```
asjohnston-asf, line 192: were the bounds intended to be `int` or `float`? we called them floats in the hyp3 PR

asjohnston-asf: asf-tools has it as an `int`, so this looks good, and we should update the type in `hyp3`
https://github.com/ASFHyP3/asf-tools/blob/develop/asf_tools/flood_map.py#L126

jacquelynsmale: yup, int. Would that just be updating '0.0' and '15.0' to '0' and '15'?

hyp3_gamma/__main__.py
```diff
@@ -183,6 +184,13 @@ def water_map():
     parser.add_argument('--hand-fraction', type=float, default=0.8)
     parser.add_argument('--membership-threshold', type=float, default=0.45)
     parser.add_argument('granule')
+
+    parser.add_argument('--include-flood-depth', type=string_is_true, default=False)
+    parser.add_argument('--estimator', type=str, default='iterative', choices=['iterative', 'logstat', 'nmad', 'numpy'])
+    parser.add_argument('--water-level-sigma', type=float, default=3., )
```
asjohnston-asf, line 190:
```suggestion
    parser.add_argument('--water-level-sigma', type=float, default=3.)
```

asjohnston-asf (approved):

PR #378, opened 2022-06-17
modify code to improve the choice of reference point

cirrusasf:


hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -97,24 +108,63 @@ def get_neighbors(array, i, j, n=1):
...
+        ref_j = cols[idx[0][0]]
+
+    return ref_i, ref_j
+
+
+def ref_point_with_max_cc(data_cc: np.array, window_size=10, pick_num=20, cc_thresh=0.3):
+    """window size determines the actual window, for example, n=1 means 9-pixel window, n=2 means 25-pixel window, etc.
+    Args:
+        data_cc: array includes cc data
+        window_size: defines the actual window, the actual window is a square of 2*window_size + 1
```
jtherrmann, line 144: I think this would be more clear.
```suggestion
    """
    Args:
        data_cc: array includes cc data
        window_size: the actual window is a square of size (2*window_size + 1); for example, window_size=1 means 9-pixel window, window_size=2 means 25-pixel window, etc.
```

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -97,24 +108,63 @@ def get_neighbors(array, i, j, n=1):
...
-def ref_point_with_max_cc(data_cc: np.array, window_size=1):
-    '''
-    shift determine the window size, n=1 9-pixel window, n=2, 25-pixel window, etc.
-    '''
+def find_ref_point_with_largest_cc(data_cc: np.ndarray, indices: (np.array, np.array),
+                                   window_size, start_idx, pick_num):
 
-    data_cc_max = data_cc[data_cc < 1.0].max()
-    idx = np.where(data_cc == data_cc_max)
+    # get the indices of the  pixels with the highest coherence values
```
jtherrmann, line 114: Should "pixel" and "coherence value" be singular here?
```suggestion
    # get the indices of the pixel with the highest coherence value

```

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -97,24 +108,63 @@ def get_neighbors(array, i, j, n=1):
     return array[i_start:i_stop, j_start:j_stop]
 
 
-def ref_point_with_max_cc(data_cc: np.array, window_size=1):
-    '''
-    shift determine the window size, n=1 9-pixel window, n=2, 25-pixel window, etc.
-    '''
+def find_ref_point_with_largest_cc(data_cc: np.ndarray, indices: (np.array, np.array),
```
jtherrmann, line 111: The name `find_ref_point_with_largest_cc` sounds very similar to `ref_point_with_max_cc`. What is the fundamental difference between those two functions?

jtherrmann: Eventually we will want some unit tests for `find_ref_point_with_largest_cc`, but we should agree on the purpose of the function first.

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -97,24 +108,63 @@ def get_neighbors(array, i, j, n=1):
...
+
+    indices = np.unravel_index(np.argsort(-data_cc, axis=None), data_cc.shape)
+
+    start_idx = 0
+
+    while True:
+
+        ref_i, ref_j = find_ref_point_with_largest_cc(data, indices, window_size, start_idx, pick_num)
+
+        if ref_i and ref_j:
```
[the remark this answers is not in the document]
cirrusasf: ref_i and ref_j can not be 0. check the find_ref_point_with_largest_cc, you will find if the ref pixel close to 0,0, we will make another round of find the suitable pixel. But to be fair, I think you idea is better. I would like to change the statement. Do you find anther needs to improve? After you done, I will modify them.

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -97,24 +108,63 @@ def get_neighbors(array, i, j, n=1):
...
     num = len(rows)
     tots = np.zeros(num, dtype=float)
 
     for k in range(num):
-        tots[k] = get_neighbors(data_cc, rows[k], cols[k], window_size).sum()
+        neighbors = get_neighbors(data_cc, rows[k], cols[k], window_size)
+        if (neighbors == 0.0).any() or neighbors.size < (2*window_size+1)**2:
+            tots[k] = 0.0
+        else:
+            tots[k] = neighbors.sum()
```
jtherrmann, line 126: Since `tots` is already filled with zeros, could you just simplify like this?
```suggestion
        if not ((neighbors == 0.0).any() or neighbors.size < (2*window_size+1)**2):
            tots[k] = neighbors.sum()
```

jtherrmann (commented):
I would suggest adding tests to https://github.com/ASFHyP3/hyp3-gamma/blob/develop/tests/insar/test_unwrapping_geocoding.py for `ref_point_with_max_cc` to demonstrate the new behavior. These tests should fail with the old version of the function, but they should pass with the new version.

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -97,28 +108,69 @@ def get_neighbors(array, i, j, n=1):
...
+    start_idx = 0
+
+    num = data_cc.size
+
+    while start_idx < num:
+
+        ref_i, ref_j = find_ref_point_from_candidates(data, indices, window_size, start_idx, pick_num)
+
+        if ref_i is not None and ref_j is not None:
+            break
```
jtherrmann, line 164: Did you want to remove this `if` statement?

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -97,28 +108,69 @@ def get_neighbors(array, i, j, n=1):
...
+
+    data[np.logical_or(data == 1.0, data < cc_thresh)] = 0.0
+
+    indices = np.unravel_index(np.argsort(-data_cc, axis=None), data_cc.shape)
+
+    start_idx = 0
+
+    num = data_cc.size
+
+    while start_idx < num:
```
jtherrmann, line 159: We can simplify this:
```suggestion
    while start_idx < data_cc.size:
```

jtherrmann:
In an effort to confirm that the modified unit test exercises the new behavior, by confirming that the test fails with the old version of `ref_point_with_max_cc`, I checked out the new version of `tests/insar/test_unwrapping_geocoding.py` on `develop`, removed all occurrences of the `pick_num` argument, and ran the tests.

I found that the second assert in `test_ref_point_with_max_cc` fails:

```python
    assert ref_point_with_max_cc(array, window_size=1) == (0, 0)
```

as does the final assertion:

```python
    assert ref_point_with_max_cc(array, window_size=1) == (4, 4)
```

Though I'm not familiar enough with the intended behavior to know how valuable this is.

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [5.4.2]
+
+### Changed
+- calculate the indices of pixel in MAP space based on the indices and height of the pixel in SAR space
+- improve the method of determination of the reference pixel in SAR space
```
asjohnston-asf, line 14:
```suggestion

### Fixed
- Map coordinates for the phase unwrapping reference point are now reported correctly in the output metadata text file
  for InSAR products. Fixes [#380](https://github.com/ASFHyP3/hyp3-gamma/issues/380)
```

asjohnston-asf, line 14:
```suggestion
### Changed
- improve the method of determination of the reference pixel in SAR space

### Fixed
- Map coordinates for the phase unwrapping reference point are now reported correctly in the output metadata text file
  for InSAR products. Fixes [#380](https://github.com/ASFHyP3/hyp3-gamma/issues/380)
```

jtherrmann, line 14: It might be nice to explain how the method has been improved.

hyp3_gamma/insar/unwrapping_geocoding.py
cirrusasf: I am thinking put data_cc.size in the while loop means that each loop, it will run data_cc.size, which waste the computing time. If it is a variable in the while loop, it does not calculate each loop, it will be faster?

hyp3_gamma/insar/unwrapping_geocoding.py
cirrusasf: I guess we need to keep the if statement, because some times we can not find the ref_i and ref_j in the, in this case, ref_i and ref_j return None, If it found the ref_i, ref_j, we end the while loop.

cirrusasf:
> In an effort to confirm that the modified unit test exercises the new behavior, by confirming that the test fails with the old version of `ref_point_with_max_cc`, I checked out the new version of `tests/insar/test_unwrapping_geocoding.py` on `develop`, removed all occurrences of the `pick_num` argument, and ran the tests.
> 
> I found that the second assert in `test_ref_point_with_max_cc` fails:
> 
> ```python
>     assert ref_point_with_max_cc(array, window_size=1) == (0, 0)
> ```
> 
> as does the final assertion:
> 
> ```python
>     assert ref_point_with_max_cc(array, window_size=1) == (4, 4)
> ```
> 
> Though I'm not familiar enough with the intended behavior to know how valuable this is.

we have to set the pick_num in the test, this argument impact the result. if we do not set the pick_num, its default value is 20. different pick_value results different ref_i and ref_j.

CHANGELOG.md
cirrusasf: where should I put the explanation?

cirrusasf:
> In an effort to confirm that the modified unit test exercises the new behavior, by confirming that the test fails with the old version of `ref_point_with_max_cc`, I checked out the new version of `tests/insar/test_unwrapping_geocoding.py` on `develop`, removed all occurrences of the `pick_num` argument, and ran the tests.
> 
> I found that the second assert in `test_ref_point_with_max_cc` fails:
> 
> ```python
>     assert ref_point_with_max_cc(array, window_size=1) == (0, 0)
> ```
> 
> as does the final assertion:
> 
> ```python
>     assert ref_point_with_max_cc(array, window_size=1) == (4, 4)
> ```
> 
> Though I'm not familiar enough with the intended behavior to know how valuable this is.

we need to set the pick_num, different pick_num produces different ref_i and ref_j.

hyp3_gamma/insar/unwrapping_geocoding.py
jtherrmann: No, `size` is just an attribute of the `data_cc` object, so it is already computed.

hyp3_gamma/insar/unwrapping_geocoding.py
jtherrmann: Ah, OK, that makes sense.

jtherrmann:
> we need to set the pick_num, different pick_num produces different ref_i and ref_j.

@cirrusasf Sorry, I did not mean to cause confusion. I was not asking for changes, I was simply documenting the steps that I took to verify that the new unit test fails with the old version of the function. My comment was only for informational purposes.

cirrusasf:
Hi Jake,

Good to know the data_cc.size thing. I will follow your suggestion.

Jiang

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -96,77 +98,26 @@ def read_bmp(file):
     return data
 
 
-def get_neighbors(array, i, j, n=1):
-    i_max, j_max = array.shape
+def sum_valid_coherence_values(array: np.array, threshold: float = 0.3):
```
jtherrmann, line 101: We already specify the default value in `get_reference_pixel`.
```suggestion
def sum_valid_coherence_values(array: np.array, threshold: float):
```

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -96,77 +98,26 @@ def read_bmp(file):
...
-        if ref_i is not None and ref_j is not None:
-            break
-
-        start_idx += pick_num
-
-    if ref_i is not None and ref_j is not None:
-        return ref_i, ref_j
-    else:
-        return 0, 0
+    data = scipy.ndimage.generic_filter(input=coherence, function=sum_valid_coherence_values, size=window_size,
```
jtherrmann, line 117: Let's rename `data`, maybe call it `sums`?

tests/insar/test_unwrapping_geocoding.py
```diff
@@ -1,49 +1,28 @@
...
-    array.shape = (5, 5)
-
-    assert np.array_equal(get_neighbors(array, 1, 1, n=2),
-                          np.array([[0, 1, 2, 3], [5, 6, 7, 8], [10, 11, 12, 13], [15, 16, 17, 18]]))
-    assert np.array_equal(get_neighbors(array, 1, 1, n=3), array)
-    assert np.array_equal(get_neighbors(array, 1, 1, n=4), array)
-
-
-def test_ref_point_with_max_cc():
+def test_get_reference_pixel():
```
jtherrmann, line 6: Do our tests cover the exclusion of `1.0` values? (At first glance it looks like probably, but it wasn't obvious.) Perhaps we should break these cases into separate functions so we can give them helpful names.

asjohnston-asf (approved):

PR #391, opened 2022-08-05
upgrade to gamma 20220630

asjohnston-asf:
Only significant change between the 202206030 linux installation guide and the one from 20210701 is the addition of `python3-packaging`. The 20220630 installation guides for ubuntu 20.04 and 22.04 are byte-for-byte identical.

I've run rtc for `S1A_IW_SLC__1SSV_20150621T120220_20150621T120232_006471_008934_72D8` and insar for `S1A_IW_SLC__1SSV_20150621T120220_20150621T120232_006471_008934_72D8+S1A_IW_SLC__1SSV_20150504T120217_20150504T120229_005771_00769E_EF9A` on my local machine with all default options. They both run to completion and produce output that looks consistent to the naked eye in a GIS.

Will need to run a full test suite to exercise all the parameter options and code paths.

PR #393, opened 2022-08-19
Gamma update 20220630 u2004

cirrusasf:


cirrusasf:
After hyp3-gamma-20220630-with-ubuntu22.04 fails the golden test. I dig in and find out the problem begin at producing big.dem/bi.par. I tested the hyp3-gamma-20220630-with-ubuntu20.04, it can produce the same big.dem/big.par as the prod, I ran the version to produce the InSAR outputs and compared them with those produced by prod.  I found the outputs are very close. 
I would like to test if the hyp3-gamma-20220630-with-ubuntu20.04 can pass golden test.

jhkennedy:
@cirrusasf #391 was already merged to `develop` making changes for updating the GAMMA version, and this PR is based on an outdated version of `develop`:
image


So you'll need to either rebase onto, or merge in, `origin/develop`. Then you'll need to look at that PR and change everything back to ubuntu 20.04, or revert that PR and then make the upgrade changes you want. 

In general, you should start each feature branch from a latest develop branch.

Dockerfile
```diff
@@ -34,7 +34,7 @@ RUN apt update \
          gdal-bin libgdal-dev \
          libhdf5-dev libhdf5-103 \
          libblas-dev libblas3 liblapack-dev liblapack3 liblapack-doc \
-         python-is-python3 python3-numpy python3-matplotlib python3-scipy python3-shapely python3-packaging \
+         python-is-python3 python3-numpy python3-matplotlib python3-scipy python3-shapely \
```
asjohnston-asf, line 37: python3-packaging is called out as required in the GAMMA 20220630 installation guide, regardless of whether we're using Ubuntu 20.04 or 22.04

asjohnston-asf:
You'll also want to revert back to Python 3.8 in test-and-build.yml at:
https://github.com/ASFHyP3/hyp3-gamma/blob/gamma-update-20220630-u2004/.github/workflows/test-and-build.yml#L25
https://github.com/ASFHyP3/hyp3-gamma/blob/gamma-update-20220630-u2004/.github/workflows/test-and-build.yml#L31

cirrusasf:
After extensive tests, I found if I separate the resample and re-projection into two steps, then gdal 3.0.4 (ubuntu20.04) and gdal 3.4.1 (ubuntu22.04) produce the same dem.tif files. In order to upgrade to gamma-20220630-u2204, we have to use the separate logic to deal with resample and re-projection. So I suggest to modify the  function prepare_dem_geotiff() in the dem.py.

hyp3_gamma/dem.py
```diff
@@ -110,5 +110,6 @@ def prepare_dem_geotiff(output_name: str, geometry: ogr.Geometry, pixel_size: fl
             gdal.BuildVRT(str(dem_vrt), dem_file_paths)
 
             epsg_code = utm_from_lon_lat(centroid.GetX(), centroid.GetY())
+
```
asjohnston-asf, line 113:
```suggestion
```

asjohnston-asf (approved):

PR #398, opened 2022-09-12
Update water mask URL

jacquelynsmale:


cirrusasf:
Using buffered method makes more water pixels into land pixels. In order to overcome this, we try to be more conservative to classify the shore line into land pixels. If this is our purpose for this update, then we should maybe classify the pixels which touch to the shoreline into water pixels.

asjohnston-asf:
That's the intent behind setting allTouched=True in call to gdal.Rasterize(), so any edge-case pixels are classified as land rather than water. I like that change!

asjohnston-asf:
@jacquelynsmale Implementation looks solid!

We should update the docstring for `create_water_mask` at to also mention Level 3, and to remove the reference to the 3km buffer.

We'll also want to remove the 3km reference from the InSAR README template at https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/metadata/templates/insar/readme.md.txt.j2#L212 and the water mask metadata template at https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/metadata/templates/insar/water_mask_tif.xml.j2#L9 . I haven't done an exhaustive search for other reference to the buffer, yet.

jacquelynsmale:
Responding to the water classification discussion: I wanted to include this image showing the classification of pixels from our test case. Transparent white represents pixels classified as land, brown is the input shapefile of land, black represents pixels classified as water, and the middle gray represents the pixels that were classified as water before the rasterize update but are classified as land with this change. 
The idea behind including any pixel that intersects with the land is to ensure we are excluding any land, especially since we know the shapefile isn't 100% accurate. 
Shapfileupdate

hyp3_gamma/metadata/templates/insar/readme.md.txt.j2
```diff
@@ -211,9 +211,9 @@ The DEM used for this product is {{ dem_name }}, which has a native pixel spacin
...
 ## 11. Water Mask
 
-The water mask identifies coastal waters and major inland waterbodies beyond 3 km of the shoreline. These areas are assigned a pixel value of 0. All remaining pixels (land, small inland waterbodies, land ice, and near-shore coastal areas) have a value of 1. The water mask is stored as an 8-bit unsigned integer GeoTIFF file and is always included in the product package.
+The water mask identifies coastal waters and major inland waterbodies. These areas are assigned a pixel value of 0. All remaining pixels (land, small inland waterbodies, and land ice) have a value of 1. The water mask is stored as an 8-bit unsigned integer GeoTIFF file and is always included in the product package.
 
-The water mask is not precise; the [GSHHG](http://www.soest.hawaii.edu/wessel/gshhg/) L1, L2 and L5 shoreline data is buffered out 3 km to reduce the possibility of near-shore features being excluded while still being a useful tool for masking out phase unwrapping artifacts over water.
+The water mask is not precise, however the [GSHHG](http://www.soest.hawaii.edu/wessel/gshhg/) L1, L2 and L5 shoreline data is  a useful tool for masking out phase unwrapping artifacts over water.
```
asjohnston-asf, line 216: Per @hjkristenson 's notes, the new shapefile now includes L3 as well, so I think we need to add it to the list here

hyp3_gamma/water_mask.py
```diff
@@ -22,14 +22,14 @@ def create_water_mask(input_tif: str, output_tif: str):
     The water mask is assembled from GSHHG v2.3.7 Levels 1, 2, and 5 at full resolution. To learn more, visit
```
asjohnston-asf, line 22: should add level 3 here

asjohnston-asf:
I see a short-form story map URL in the README, but a long-form URL everywhere else. I'd consider using the short-form everywhere, assuming it's as long-lived as the longer URL?
https://arcg.is/19aKHj
https://storymaps.arcgis.com/stories/485916be1b1d46889aa436794b5633cb

asjohnston-asf (approved):

asjohnston-asf:
I've reviewed the secret analysis workflow and confirmed it's a false-positive for the story map URLs. Fine to merge.

PR #400, opened 2022-09-21
Update base Ubuntu image to 22.04 for docker

jhkennedy:


jhkennedy:
trufflehog is complaining about ArcGIS story map links, so should be fine to ignore for now and will clean up next release

asjohnston-asf (approved):

PR #401, opened 2022-09-23
set OGR_GEOJSON_MAX_OBJ_SIZE in create_water_map

asjohnston-asf:
Here's a crude and quick fix to the `ERROR 1: GeoJSON object too complex, please see the OGR_GEOJSON_MAX_OBJ_SIZE environment option` errors we see when processing these scenes over the Aegean Sea:
- `S1B_IW_SLC__1SDV_20201115T162313_20201115T162340_024278_02E29D_5C54`
- `S1A_IW_SLC__1SDV_20201203T162353_20201203T162420_035524_042744_6D5C`

The `OGR_GEOJSON_MAX_OBJ_SIZE` setting is specific to the GDAL GeoJSON driver, documentation is at https://gdal.org/drivers/vector/geojson.html#configuration-options

We're using GeoJSON for the temp file out of convenience, we might be able to avoid this issue by using any other [GDAL-supported vector format](https://gdal.org/drivers/vector/index.html).

Alternatively, if we stick with GeoJSON, it would be nice to set `OGR_GEOJSON_MAX_OBJ_SIZE` based on the size of the temporary geojson file instead of to an arbitrary big value.

@jacquelynsmale any thoughts on those options? Would you be interested in taking a stab at implementing either? I've only checked that this code still passes the unit tests, I haven't yet verified it fixes the issue for the specific pair in question.

### DIFF
--- a/hyp3_gamma/water_mask.py
+++ b/hyp3_gamma/water_mask.py
@@ -6,6 +6,8 @@ from tempfile import NamedTemporaryFile
 import geopandas
 from osgeo import gdal
 
+from hyp3_gamma.util import GDALConfigManager
+
 gdal.UseExceptions()
 
 
@@ -44,6 +46,7 @@ def create_water_mask(input_tif: str, output_tif: str):
     mask = geopandas.read_file(mask_location, mask=extent)
     with NamedTemporaryFile() as temp_file:
         mask.to_file(temp_file.name, driver='GeoJSON')
-        gdal.Rasterize(dst_ds, temp_file.name, allTouched=True, burnValues=[1])
+        with GDALConfigManager(OGR_GEOJSON_MAX_OBJ_SIZE='500MB'):
+            gdal.Rasterize(dst_ds, temp_file.name, allTouched=True, burnValues=[1])
 
     del src_ds, dst_ds
--- a/hyp3_gamma/water_mask.py
+++ b/hyp3_gamma/water_mask.py
@@ -1,7 +1,7 @@
 """Create and apply a water body mask"""
 import json
 import subprocess
-from tempfile import NamedTemporaryFile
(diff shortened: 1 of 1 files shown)


hjkristenson:
@asjohnston-asf Is there any reason why we don't just use the ESRI-shapefile format? That's what the shorelines file is, and I see that it's supported in the list that you linked. Is it just that the extracted portion of the shapefile is rendered in GeoJSON right now? Is it much harder to just keep it in the shapefile format?

asjohnston-asf:
That was my first thought, my only (very minor) hesitation is that format is multiple files, so we might have to tweak the temp file details to clean up the temporary sidecar files.

hjkristenson:
That makes sense. I guess we could test the bumped-up GeoJSON approach to see if it does the trick, and try out the shapefile approach if there are issues (or concerns about what it would take to find an appropriate object size limit).

hyp3_gamma/water_mask.py
```diff
@@ -44,6 +46,7 @@ def create_water_mask(input_tif: str, output_tif: str):
     mask = geopandas.read_file(mask_location, mask=extent)
     with NamedTemporaryFile() as temp_file:
         mask.to_file(temp_file.name, driver='GeoJSON')
```
jacquelynsmale, line 48: The number of bytes can be returned with `os.stat(temp_file.name).st_size`, so guess if we update OGR_GEOJSON_MAX_OBJ_SIZE=(os.stat(temp_file.name.st_size)+(buffer number))/1000000, that could work? It's a bit clunky, though

asjohnston-asf:
I've confirmed the trufflehog failure is a false-positive related to story map URLs already merged to develop.

jacquelynsmale (approved):

PR #416, opened 2022-11-11
Update asf-tools to v0.4.3

jacquelynsmale:


jhkennedy (approved):
I think for testing this, we could run a watermap job in the HKH region in hyp3-watermap and hyp3-test, and we should see a significant change in the included HAND tile.

PR #428, opened 2022-12-31
add ++omp-num-threads parameter to main() entrypoint

asjohnston-asf:
GAMMA programs use OpenMP for parallel processing. The number of threads used range from 3-12+ from program to program. This can lead to severe CPU contention when running multiple GAMMA jobs on the same machine, drastically increasing the run time of those jobs.

This PR adds a command line parameter to the `main()` entrypoint to set the `OMP_NUM_THREADS` environment variable. See https://www.ibm.com/docs/en/zos/2.2.0?topic=suce-environment-variables-openmp

This allows running multiple jobs on the same machine without CPU contention. For example, running two jobs on an 8 core machine would typically result in both jobs trying to use all 8 cores at the same time. Setting `++omp-num-threads=4` for each job would limit the total demand to match the number of available cores and eliminates the performance degradation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [5.7.4]
+
+### Added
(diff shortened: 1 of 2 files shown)


jhkennedy (approved):
Ooh I like this approach

PR #429, opened 2023-01-05
Add choice of 10m to rtc_sentinel

jacquelynsmale:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ### Added
 - `++omp-num-threads` parameter to the `main()` entrypoint to limit the number of threads used by GAMMA during
   multiprocessing.
+- Choice 10 m to the `resolution` parameter for `hyp3_gamma.rtc_sentinel`
 
 ## [5.7.3]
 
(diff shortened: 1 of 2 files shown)


jhkennedy (commented):
I'm not sure this PR is needed at all? hyp3-gamma already runs 10 RTC jobs in the `hyp3-streamflow` deployment, and this particular entrypoint is a "science" one which *should* be able to accept any resolution (we may want to try out different resolutions on our own). Adding `choices` here restricts this function to *only* 10m and 30m resolutions.

jacquelynsmale:
> I'm not sure this PR is needed at all? hyp3-gamma already runs 10 RTC jobs in the `hyp3-streamflow` deployment, and this particular entrypoint is a "science" one which *should* be able to accept any resolution (we may want to try out different resolutions on our own). Adding `choices` here restricts this function to *only* 10m and 30m resolutions. 
> 
> Plus, they HyP3 entrypoint already has the acceptable choices for hyp3:
> https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/__main__.py#L68

Yea, that makes sense. I mostly got the ball rolling on this because it was 5pm and I was trying to address https://asfdaac.atlassian.net/browse/TOOL-1257 ;) I think that's a really good point, though. Would we want to restrict this? If not, we can definitely close this PR

jhkennedy:
@jacquelynsmale ah! TOOL_1257 is for [hyp3](https://github.com/ASFHyP3/hyp3) not hyp3-gamma. :-D

PR #435, opened 2023-01-17
drop asf-tools env from container and watermap hyp3 entrypoint

jhkennedy:
Once HyP3 uses asf-tools directly to run the water map and flood depth code, we can drop all that from this container. 

Will need to test that RTC and InSAR jobs still work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [5.8.0]
+
+### Removed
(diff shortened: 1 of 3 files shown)


forrestfwilliams:
@jhkennedy The changes here look good, nothing other than the portions pertaining to `asf-tools` has been dropped. I'd like to try rebuilding the container locally and running InSAR/RTC jobs before giving the thumbs-up however. I'll likely get to this Monday morning.

jhkennedy:
@forrestfwilliams you can also merge and try it in test if you like. Guess it'll need a changelog first

forrestfwilliams:
That works for me. If you add a changelog entry, I'll approve.

jhkennedy:
@forrestfwilliams done!

forrestfwilliams (approved):
LGTM!

PR #441, opened 2023-02-14
Fix Copernicus DEM links

jtherrmann:
This is a start on fixing the broken Copernicus DEM links. So far the links should be fixed in the `insar` and `rtc` README templates as well as the macros file.

However, if we grep the repo for the first part of the URL (note that this matches both the corrected links and the broken links because they're under the same domain name `spacedata.copernicus.eu`):

```
git grep -n spacedata
```

We find the broken links in several other files as well:
- [x] `hyp3_gamma/metadata/templates/dem/dem-cop.xml.j2`
- [x] `hyp3_gamma/metadata/templates/insar/dem.xml.j2`
- [x] `hyp3_gamma/metadata/templates/insar/inc_map_tif.xml.j2`
- [x] `hyp3_gamma/metadata/templates/insar/insar_dem.xml.j2`

After fixing the remaining links, we should then generate example metadata for both `rtc` and `insar` as described [here](https://github.com/ASFHyP3/hyp3-gamma/tree/develop/hyp3_gamma/metadata#readme) and confirm that the links have been fixed (`grep` the example products for `spacedata`).

Note that I have elected to replace broken links with `https://spacedata.copernicus.eu/collections/copernicus-digital-elevation-model` except for links to the product handbook, which I am simply removing. See [my edits so far](https://github.com/ASFHyP3/hyp3-gamma/pull/441/files).


forrestfwilliams:
@jtherrmann PR #442 fixes the files you indicated with your strategy. Feel free to merge into this branch if the changes look good.

jtherrmann:
After running `git grep -n spacedata` again and performing a visual spot check of all the links, it appears that they have all been updated to `https://spacedata.copernicus.eu/collections/copernicus-digital-elevation-model`.

jtherrmann:
I ran `git grep -ni handbook` before and after Forrest's changes, and confirmed that there were several matches before, and none after.

jtherrmann:
I generated the example metadata files for `rtc` and `insar` and ran `grep spacedata -nr .` in each product directory, and confirmed that all the matches are the correct link.

jhkennedy (approved):
Looks like it covers all the links I could find.

PR #442, opened 2023-02-14
Update links in Jinja templates

forrestfwilliams:


jhkennedy (approved):

jtherrmann (commented):
I know this has already been merged, but I confirmed that the added lines contain only the correct link, and all mentions of the product handbook are present only in the deleted lines.

PR #447, opened 2023-03-31
Adjustable adf avo

cirrusasf:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [6.0.2]
 
 ### Added
-- `phase_filter_threshold` option for the processing of INSAR products 
+- A `--phase-filter-paameter` option has been added to the `__main__` (HyP3) and `ifm_sentinel.py` entry points to specify the adaptive phase filter parameter used when processing InSAR products. The provided value is ultimately passed to the `alpha` argument of the `adf` GAMMA function, which uses the algorithm with a constant exponent as described in https://doi.org/10.1029/1998GL900033.
 
 ## [6.0.1]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [6.0.2]
+
+### Added
+- `phase-filter-threshold` option for the processing of INSAR products 
+
 ## [6.0.1]
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 ## [6.0.2]
 
 ### Added
-- `phase-filter-threshold` option for the processing of INSAR products 
+- `phase_filter_threshold` option for the processing of INSAR products 
 
 ## [6.0.1]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [6.0.2]
+## [6.1.0]
 
 ### Added
 - A `--phase-filter-paameter` option has been added to the `__main__` (HyP3) and `ifm_sentinel.py` entry points to specify the adaptive phase filter parameter used when processing InSAR products. The provided value is ultimately passed to the `alpha` argument of the `adf` GAMMA function, which uses the algorithm with a constant exponent as described in https://doi.org/10.1029/1998GL900033.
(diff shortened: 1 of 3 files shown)


hyp3_gamma/__main__.py
```diff
@@ -112,6 +112,17 @@ def rtc():
...
 
+def valid_range_float(x):
+    try:
+        x = float(x)
+    except ValueError:
+        raise ArgumentTypeError("%r not a floating-point literal" % (x,))
+
+    if x < 0.0 or x > 1.0:
+        raise ArgumentTypeError("%r not in range [0.0, 1.0]" % (x,))
+    return x
```
jhkennedy, line 123:
```suggestion
def valid_range_float(x):
    x = float(x)
    if 0.0 <= x <= 1.0:
        return x
    raise ValueError(f'{x} not in range [0.0, 1.0]')


```

For string formatting, you should always prefer f-strings over `.format` strings over `%`. 

You also do not need to raise an `ArgumentTypeError` here, argparse will do that for you automatically when specifying this function as a type.

hyp3_gamma/__main__.py
```diff
@@ -126,6 +137,7 @@ def insar():
     parser.add_argument('--include-inc-map', type=string_is_true, default=False)
     parser.add_argument('--apply-water-mask', type=string_is_true, default=False)
     parser.add_argument('--looks', choices=['20x4', '10x2'], default='20x4')
+    parser.add_argument('--phase-filter-threshold', type=valid_range_float, default=0.6)
```
jhkennedy, line 140: It'd be worth checking the GAMMA docs to see what they call this parameter.

jhkennedy (commented):
@cirrusasf do you think we should also expose the `width` parameter? I'm leaning towards yes.
https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/insar/unwrapping_geocoding.py#L240

hyp3_gamma/__main__.py
cirrusasf: I will check the width in the GAMMA.

cirrusasf:
> @cirrusasf do you think we should also expose the `width` parameter? I'm leaning towards yes. https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/insar/unwrapping_geocoding.py#L240

From my understanding, width is called interferogram_width, it is obtained from the off.it file, and it is not changeable parameter. So maybe we should not change this.

jhkennedy:
> I checked the code as well as GAMMA doc, it looks like the width is called interferogram_width, it is obtained from the off.it file, and it is not a changeable parameter. So maybe we should not change this.

Perfect. Then yes, I'd agree.

hyp3_gamma/insar/ifm_sentinel.py
```diff
@@ -484,7 +487,7 @@ def main():
     insar_sentinel_gamma(args.reference, args.secondary, rlooks=args.rlooks, alooks=args.alooks,
                          include_look_vectors=args.l, include_displacement_maps=args.s,
                          include_wrapped_phase=args.w, include_inc_map=args.i,
-                         include_dem=args.d, apply_water_mask=args.m)
+                         include_dem=args.d, apply_water_mask=args.m, phase_filter_thrshold=args.t)
```
jhkennedy, line 490:
```suggestion
                         include_dem=args.d, apply_water_mask=args.m, phase_filter_thrshold=args.phase_filter_threshold)
```
Argparse will use the long name if provided.

cirrusasf:
> > I checked the code as well as GAMMA doc, it looks like the width is called interferogram_width, it is obtained from the off.it file, and it is not a changeable parameter. So maybe we should not change this.
> 
> Perfect. Then yes, I'd agree.

Good!

hyp3_gamma/__main__.py
jhkennedy: Looking at the GAMMA docs, this is the exponent in the phase filter. I'm unsure what the valid range is now, but the algorithm is described in this paper:
https://doi.org/10.1029/1998GL900033

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [6.0.2]
+
+### Added
+- `phase_filter_threshold` option for the processing of INSAR products
```
jhkennedy, line 12:
```suggestion
- A `--phase-filter-paameter` option has been added to the `__main__` (HyP3) and `ifm_sentinel.py` entry points to specify the adaptive phase filter parameter used when processing InSAR products. The provided value is ultimately passed to the `alpha` argument of the `adf` GAMMA function, which uses the algorithm with a constant exponent as described in https://doi.org/10.1029/1998GL900033.
```

jhkennedy (changes requested):
Since the manuscript the algorithm is adapted from, and our parameter file, calls it a phase filter parameter, it's worth using the same name. 

I've also included a lot of the details in the changelog so it's easy to trace back if/when we add documentation for this parameter to hyp3-docs

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [PEP 440](https://www.python.org/dev/peps/pep-0440/)
 and uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [6.0.2]
```
jhkennedy, line 9:
```suggestion
## [6.1.0]
```
This is a minor enhancement as we're adding functionality.

jhkennedy (approved):
Thanks for applying the changes -- looks good to me!

PR #452, opened 2023-05-02
Release V6.2.0

cirrusasf:
### Developer checklist

- [x] Indicated the level of changes to this package by affixing one of these labels:
  * major -- Major changes to the API that may break current workflows
  * minor -- Minor changes to the API that do not break current workflows
  * patch -- Patches and bugfixes for the current version that do not break current workflows
  * bumpless -- Changes to documentation, CI/CD pipelines, etc. that don't affect the software's version
- [x] (If applicable) Updated the dependencies and indicated any downstream changes that are required
- [x] Added/updated documentation for these changes
- [x] Added/updated tests for these changes
- [ ] If the step function code has changed, have you drained the job queue before merging?

  * For example, if the interface for a Lambda function has changed to expect different input,
    then currently running jobs (which use the old step function definition) will call the new
    function with the old input. So we must drain the job queue before deployment, so that the new
    function is only called by the new step function definition.

### Reviewer checklist

- [x] Have all dependencies been updated?
- [x] Is the level of changes labeled appropriately?
- [x] Are all the changes described appropriately in `CHANGELOG.md`?
- [x] Has the documentation been adequately updated?
- [x] Are the tests adequate?

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -13,6 +13,6 @@ on:
 
 jobs:
   call-changelog-check-workflow:
-    uses: ASFHyP3/actions/.github/workflows/reusable-changelog-check.yml@v0.7.0
+    uses: ASFHyP3/actions/.github/workflows/reusable-changelog-check.yml@v0.7.1
     secrets:
       USER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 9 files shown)


AndrewPlayer3 (approved):
Looks good to me!

Issue #260, opened 2021-05-13
Output DEM included with RTC/InSAR products potentially has invalid data in areas where DEM data was not needed for processing


asjohnston-asf:
The DEM included via the `--include-dem` option for InSAR and RTC jobs has the same spatial extent as the other output products.  However, the DEM geotiff is built from only DEM tiles that overlap areas of data, and therefore can have invalid data in the corners where DEM data was not needed for underlying processing.

This example is for S1A_IW_SLC__1SDH_20200307T133427_20200307T133455_031570_03A324_C47F .  The output DEM (pictured with the backscatter geotiff overlayed) has an area of invalid data in the upper left of the image.

![Screenshot from 2021-05-13 10-34-46](https://user-images.githubusercontent.com/17994518/118171033-50362b80-b3d7-11eb-871c-98603b89c0e6.png)

This issue does not impact RTC/InSAR processing, but has the potential to impact workflows that use the DEM included with the output products.

asjohnston-asf:
This is potentially resolved by fetching all DEM tiles that intersect the envelope of the scene's geometry (instead of the geometry itself) at https://github.com/ASFHyP3/hyp3-gamma/blob/develop/hyp3_gamma/dem.py#L103

The naive implementation using `ogr.Geometry.GetEnvelope()` will have issues with the antimeridian, however.

asjohnston-asf:
@cirrusasf notes that these invalid data in the DEM also create corresponding artifacts in the local incidence angle map.
![dem](https://user-images.githubusercontent.com/17994518/127217945-e74e772f-9d3f-4eef-b625-c277c76d672f.png)
![inc_map](https://user-images.githubusercontent.com/17994518/127217954-60be249e-4e5b-48e6-a119-f455bd043189.png)

jhkennedy:
To illustrate:
![image](https://user-images.githubusercontent.com/7882693/180900542-adee391d-49a5-4544-8433-76a4cf952ca7.png)

[...]

PR #464, opened 2023-06-22
Change nodata in rtc dem

cirrusasf:
See https://github.com/ASFHyP3/hyp3-gamma/issues/260


hyp3_gamma/dem.py
```diff
@@ -99,11 +119,14 @@ def prepare_dem_geotiff(output_name: str, geometry: ogr.Geometry, pixel_size: fl
         with TemporaryDirectory() as temp_dir:
             temp_path = Path(temp_dir)
 
+            geometry = get_envelope_geometry(geometry)
```
jtherrmann, line 122: Why can't we just do the following?

```python
geometry = geometry.GetEnvelope()
```

cirrusasf: For geometry does not cross over anti meridian, we use  geometry = geometry.GetEnvelope(), but this does not work for gemoetry that croses over anti-meridian, This is why I developed the function for the purpose.

jtherrmann: That makes sense, thanks for the clarification. Can you add some unit tests for this function? We should probably have one test for a geometry that does not cross the anti-meridian, and another test for a geometry that crosses the anti-meridian.

cirrusasf:
Hi Jack,
I will add the  unit test for the function, thank you!

Jiang

cirrusasf:
We decide not to create the input dem.tif based on the envelope of map_overlay, only modify the code to produce the  output dem.tif without setting nodata. New code passes local test. It is ready for review.

PR #484, opened 2023-09-22
Release v6.3.1

asjohnston-asf:


jtherrmann (approved):

hyp3_gamma/insar/unwrapping_geocoding.py
```diff
@@ -201,7 +201,7 @@ def combine_water_mask(cc_mask_file, water_mask_sar_file):
 
 
 def unwrapping_geocoding(reference, secondary, step="man", rlooks=10, alooks=2, trimode=0,
-                         npatr=1, npata=1, alpha=0.6, apply_water_mask=False):
```
jhkennedy, line 204: Notably, this removes the ability of the `unwrapping_geocoding` entrypoint ("science code") to process interferograms with different range/azimuth numbers of patches. 

However, I'm not sure anyone's really using those options, so I'm OK with this, but I think it might be worth noting explicitly in the changelog. I'm going to approve either way, I think, so take it or leave it.

jhkennedy (approved):

PR #503, opened 2023-10-16
Copy pre-ADF files

jtherrmann:


forrestfwilliams (approved):
These changes LGTM. They match the workflow we discussed and all the tests have passed.

PR #512, opened 2023-11-01
need to add libspatialindex-dev and rtree for making the geopandas.cl…

cirrusasf:
…ip() work


environment.yml
```diff
@@ -30,4 +30,5 @@ dependencies:
   - pip:
     - lxml==4.8.0
     - hyp3lib==1.7.0
-    - pyarrow
+    - rtree
+    - pyarrow
```
jhkennedy, line 34: We pip install lxml and hyp3lib due to a conda build issue (see comment in env file) but I don't see why are we installing rtree and pyarrow via pip instead of conda (they are both available via conda)

cirrusasf: Ok, I will move them to conda install portion

jhkennedy (approved):

Dockerfile
```diff
@@ -33,7 +33,7 @@ RUN apt update \
          python-is-python3 \
          python3-numpy python3-matplotlib python3-scipy python3-shapely python3-packaging \
          # Additional installs
-         python3-pip wget git vim \
+         python3-pip wget git vim libspatialindex-dev \
```
jhkennedy, line 36: @cirrusasf are you sure we need this? the `rtree` python package page implies that it bundles `libspatialindex` in the wheel and we should only need to pip install `rtree`:
image
https://pypi.org/project/Rtree/

And in the docker build, I see a wheel being installed:
https://github.com/ASFHyP3/hyp3-gamma/actions/runs/6727289790/job/18284909084#step:8:3611

cirrusasf: Joe, since you have already approved it. Does it mean we still keep the libspatialindex-dev in the Dockerfile and use conda to install rtree? I have merged the PR. If we need otherwise, please let me know.

PR #514, opened 2023-11-02
back to the previous way of reasding the mask with geopandas.read_fil…

cirrusasf:
…e, add clip with extent

### DIFF
--- a/hyp3_gamma/water_mask.py
+++ b/hyp3_gamma/water_mask.py
@@ -103,12 +103,12 @@ def create_water_mask(input_image: str, output_image: str, gdal_format='GTiff'):
     dst_ds.SetMetadataItem('AREA_OR_POINT', src_ds.GetMetadataItem('AREA_OR_POINT'))
 
     extent = gdal.Info(input_image, format='json')['wgs84Extent']
-    extent = split_geometry_on_antimeridian(extent)
+    corrected_extent = split_geometry_on_antimeridian(extent)
     # mask = get_water_mask_gdf(extent)
 
     mask_location = '/vsicurl/https://asf-dem-west.s3.amazonaws.com/WATER_MASK/GSHHG/hyp3_water_mask_20220912.shp'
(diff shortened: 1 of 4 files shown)


cirrusasf:
I tested the docker image for this version locally. It produces correct insar product for the anti-meridian pair: S1A_IW_SLC__1SDV_20230608T183205_20230608T183234_048898_05E15B_663F
S1A_IW_SLC__1SDV_20230702T183206_20230702T183236_049248_05EC01_E364

This version also removed the  libspatial-dev in the Dockerfile. The only problem is this version runs slowly in loading the GSHHG shapefile. If we want to use parguet, we need to modify some of the get_water_mask_gdf

hyp3_gamma/water_mask.py
```diff
@@ -43,6 +45,37 @@ def split_geometry_on_antimeridian(geometry: dict):
...
+    Args:
+        box: [min_lon, min_lat, max_lon, max_lat]
+    Returns:
+        polygon
+    '''
+    p1 = geometry.Point(box[0], box[1])
+    p2 = geometry.Point(box[2], box[1])
+    p3 = geometry.Point(box[2], box[3])
+    p4 = geometry.Point(box[0], box[3])
+    return geometry.Polygon([p1, p2, p3, p4, p1])
```
AndrewPlayer3, line 60: An additional point was at the end of this list, and I think it's a bit clearer like this.
```suggestion
    min_lon = geometry.Point(box[0], box[1])
    min_lat = geometry.Point(box[2], box[1])
    max_lon = geometry.Point(box[2], box[3])
    max_lat = geometry.Point(box[0], box[3])
    return geometry.Polygon([min_lon, min_lat, max_lon, max_lat])
```

AndrewPlayer3: Other than that, everything looks good. I verified for correctness of the masks in a local environment.

AndrewPlayer3 (approved):

PR #518, opened 2023-11-08
Removed Unnecessary Imports

AndrewPlayer3:


cirrusasf (commented):
you can delete s3fs, it is reguired by parquet, you do need rtree, it is required by geopandas.clip(), I forgot which need the pyarrow

cirrusasf (commented):
I will figure out if we need pyarrow

cirrusasf (changes requested):
keep rtree and delete s3fs and pyarrow in both environment.yml and step.py,  The rtree is required by geopandas.clip(), so we must keep it. S3fs and pyarrow are required by parquet, since we do not use parquet any more, we can delete these two libraries.

cirrusasf (approved):

cirrusasf:
I changed the environment.yml and setup.py and approved the PR. You can go ahead to merge it now.

PR #530, opened 2023-11-22
Revise water mask

cirrusasf:


AndrewPlayer3 (approved):
I tested this change locally, and it seems to work great.

PR #549, opened 2024-02-20
Remove GSHHG reference

forrestfwilliams:


hjkristenson (approved):

hyp3_gamma/metadata/templates/insar/readme.md.txt.j2
```diff
@@ -228,7 +228,7 @@ The DEM used for this product is {{ dem_name }}, which has a native pixel spacin
 
 The water mask identifies coastal waters and major inland waterbodies. These areas are assigned a pixel value of 0. All remaining pixels (land, islands in large lakes, small inland waterbodies, and landfast Antarctic ice) have a value of 1. The water mask is stored as an 8-bit unsigned integer GeoTIFF file and is always included in the product package.
```
hjkristenson, line 229: Is this still accurate for the new mask? It seems like the new mask is much more aggressive with the small lakes (we didn't bother masking small lakes with the GSHHG; it was just the biggest inland waterbodies), and I'm not sure what the approach for landfast Antarctic ice is with the new datasets. @AndrewPlayer3 might have more insight on that.

forrestfwilliams: I guess it hangs on what you define as "major". I'm inclined to keep the language the same since Andrew did try to include comparable waterbodies. Luckily we'll be publishing the code for this new water mask, so the truly curious will be able to determine exactly what we did.

jacquelynsmale (approved):

PR #574, opened 2024-07-02
Upgrade to GAMMA 20231208

forrestfwilliams:
Notable changes:
- *dem_gradient_vec*: Calculate the gradient vector or the normal surface vector in ENU coordinates.
The unit surface gradient vector lies on the surface and points in the direction of the steepest
descent. The normal vector is perpendicular to the tangent plane.
- *phase_sim_orb, phase_sim*: Improved accuracy in the calculation of the simulated phase. Default
value for [ph_mode] has now been set to "1: subtract phase offset at scene center to improve
precision"
- *geocoding.py*: Now also supports using a constant height value instead of a DEM file.

I don't see any changes that should negatively impact our processing.


asjohnston-asf (approved):

asjohnston-asf:
Approved, but converting to a draft as a signal to avoid merging just yet. We have a handful of python upgrades currently in test. @jtherrmann and I are going to get those validated and released to production so we can have _only_ the GAMMA upgrade in development during testing.

forrestfwilliams:
Sounds good @asjohnston-asf, I'll hold off until I hear from you all. FYI, golden tests passed!

PR #581, opened 2024-07-11
Update to gamma 20240627 instead

forrestfwilliams:


jtherrmann (approved):

.github/workflows/test-and-build.yml
```diff
@@ -58,8 +58,8 @@ jobs:
 
       - name: Fetch GAMMA
         run: |
-          aws s3 cp s3://hyp3-software/GAMMA_SOFTWARE-20231208_ISP_DIFF_LAT.linux64_ubuntu2204.tar.gz .
-          tar -zxvfGAMMA_SOFTWARE-20231208_ISP_DIFF_LAT.linux64_ubuntu2204.tar.gz
+          aws s3 cp s3://hyp3-software/GAMMA_SOFTWARE-20240627_MSP_ISP_DIFF_LAT.linux64_ubuntu2404.tar.gz .
+          tar -zxvf GAMMA_SOFTWARE-20240627_MSP_ISP_DIFF_LAT.linux64_ubuntu2404.tar.gz
```
asjohnston-asf, line 62: Our Dockerfile is ubuntu 22.04; we should use the 22.04 installer for GAMMA, not 24.04.

PR #600, opened 2025-01-10
Add `mypy` Action for Type Checking

AndrewPlayer3:


pyproject.toml
```diff
@@ -77,3 +77,12 @@ convention = "google"
...
 lines-after-imports = 2
+
+[tool.mypy]
+python_version = "3.10"
+warn_redundant_casts = true
+warn_unused_ignores = true
+warn_unreachable = true
+strict_equality = true
+check_untyped_defs = true
+explicit_package_bases = true
```
jtherrmann, line 88: What does `explicit_package_bases` do?

requirements.txt
```diff
@@ -16,3 +16,4 @@ pyproj==3.7.0
 python-dateutil==2.9.0.post0
 rtree==1.3.0
 s1_orbits==0.1.1
+mypy
```
jtherrmann, line 19: Let's move `mypy` to `requirements-develop.txt`, which gets installed with the mamba environment.
```suggestion
```

tests/test_water_mask.py
```diff
@@ -78,6 +78,4 @@ def test_create_water_mask(tmp_path, test_data_dir):
     info_from_img = gdal.Info(output_image)
     info_from_txt = open(validation_text).read()
     # The first 4 lines are file paths; they will be different everytime and don't need to be checked.
-    info_from_img = info_from_img.split('\n')[4:]
-    info_from_txt = info_from_txt.split('\n')[4:]
-    assert info_from_img == info_from_txt
+    assert info_from_img.split('\n')[4:] == info_from_txt.split('\n')[4:]
```
jtherrmann, line 81: Was this causing a `mypy` error? Also, this comment seems unnecessary and is also incorrect (I think it should say `The first 3 lines`), should we just remove it?

pyproject.toml
jtherrmann: Documentation is here, looks like it's working as expected: https://mypy.readthedocs.io/en/stable/running_mypy.html#mapping-file-paths-to-modules

AndrewPlayer3: There are 2 `conftest.py` files, which results in a `mypy` error. This fixes that per the suggestion in the error.

tests/metadata/test_rtc_gamma.py
```diff
@@ -43,10 +43,12 @@ def test_rtc_gamma_product(rtc_product_dir):
         rtc_product_dir / 'S1A_IW_20150621T120220_DVP_RTC10_G_saufem_F8E2_VH.tif.xml',
     ]
     for output_file in output_file_list:
+        assert output_file is not None
         assert output_file.exists()
 
 
 def test_create_dem_xml(rtc_product_dir):
+    dem_name: str | None = None
```
jtherrmann, line 51: You actually don't need to assign a value, it can just be a standalone type annotation.
```suggestion
    dem_name: str | None
```

jtherrmann (approved):

PR #628, opened 2025-04-16
Release v9.0.3

asjohnston-asf:
InSAR: https://github.com/ASFHyP3/hyp3-testing/actions/runs/14500416144
RTC: https://github.com/ASFHyP3/hyp3-testing/actions/runs/14500408824

Both failed as expected. Reviewing a handful of outputs from each I'm content to call them "close enough". The only significant change is that output geotiffs are no longer clipped to the available COP30 DEM tiles, as noted in the changelog.


hyp3_gamma/dem.py
```diff
@@ -1,98 +1,37 @@
...
-
-def get_dem_file_paths(geometry: ogr.Geometry) -> list[str]:
-    file_paths = []
-    for feature in get_dem_features():
-        if feature.GetGeometryRef().Intersects(geometry):
-            file_paths.append(feature.GetField('file_path'))
-    return file_paths
-
-
 def utm_from_lon_lat(lon: float, lat: float) -> int:
```
forrestfwilliams, line 28: Eventually I think this function would be good to move this to hyp3-lib because I think more than one plugin has functionality like this. I'm not advocating we do it now though.

forrestfwilliams (approved):
I've also independently verified the DEM changes and am confident we are not making scientifically significant changes.

PR #645, opened 2025-05-22
Release v9.0.6

jacquelynsmale:
Release to fix granules getting clipped at extremely northern and southern latitudes by DEM buffer.


williamh890 (approved):
One thing we might want to keep an eye on is whether or not to move the distance buffering/rounding of the buffer into `hyp3-lib`. I think it's fine for now, but if this pops up agains it's something to consider.

PR #701, opened 2026-06-03
Streamline S1 mission descriptions

hjkristenson:


hjkristenson:
The diffs don't handle the xml content changes very well, but here's the new version as rendered in ArcGIS: 
image

This way we won't have to revisit the content once S1A is decommissioned (or for any other changes).

yoreley777 (approved):

